PR #3, opened 2020-05-18
Remove bundled libraries

sfan5:


nerzhul:
Fine to me, it will prevent bad linking

PR #5, opened 2020-05-19
Fix loaded texture log spam

appgurueu:
Should close minetest/minetest/issues/9716, but not sure


ClobberXD:
Thank. you. so. much... :)

PR #9, opened 2020-12-31
Rewrite engine to use SDL only [WIP]

nerzhul:


source/Irrlicht/Irrlicht.cpp
```diff
@@ -64,53 +64,7 @@ namespace irr
...
-		if (dev && !dev->getVideoDriver() && params.DriverType != video::EDT_NULL)
-		{
-			dev->closeDevice(); // destroy window
-			dev->run(); // consume quit message
-			dev->drop();
-			dev = 0;
-		}
-
-		return dev;
+		return new CIrrDeviceLinux(params);
```
sfan5, line 67: This seems to mix the (old) Linux/X11 device with all new SDL code.

I have doubt that this works on Windows or macOS.

nerzhul: it's not ready yet, just here for the posterity as a startup :)

sfan5: What I'm trying to say is: it might be better to create a new CIrrDeviceSDL2 then

nerzhul: yes it should be a good option, i will try to take more time to make it work properly :)

nerzhul: it's done now, linux device is dropped. I need to rebind events. hopfully there is a SDL1 backend to help us to translate some orders from WM system, but not all :)

nerzhul:
works a bit, keymap is currently broken and mouse is not show on screen. Didn't tested android yet, sounds to be handled by SDL itself and can be improved in irrlicht internals directly

PR #10, opened 2021-01-16
SDL integration improvements

numberZero:
Have no write access, so here are my commits. More is to be pushed.


numberZero:
As SDL supports all the necessary target platforms, I removed other “Devices” (not to confuse with “Video Drivers”).

.github/workflows/build.yml
```diff
@@ -13,10 +13,25 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install deps
         run: |
-          sudo apt-get install g++ gcc libxxf86vm-dev libgl1-mesa-dev libbz2-dev libpng-dev libjpeg-dev zlib1g-dev -qyy
+          sudo apt-get install -qyy cmake g++ gcc libbz2-dev libpng-dev libjpeg-dev wget zlib1g-dev \
+            libasound2-dev libdbus-1-dev libegl1-mesa-dev libgl1-mesa-dev libgles2-mesa-dev libglu1-mesa-dev libibus-1.0-dev libpulse-dev libsndio-dev libudev-dev libwayland-dev libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxkbcommon-dev libxrandr-dev libxss-dev libxt-dev libxv-dev libxxf86vm-dev 
+
+      - name: Build SDL
+        run: |
+          wget https://www.libsdl.org/release/SDL2-2.0.14.tar.gz
```
sfan5, line 21: Uh, why not just install it from the repos?

numberZero: Because SDL 2.0.14 is required (dunno why, that might be relaxed later) while Ubuntu 18.04 has only 2.0.12

sfan5:
I think it'd be better to either have separate PRs for the device removal + build system change* _or_ squash commits appropriately.
\* I don't mind that btw, it just needs to work fine for the platforms we support (and specifically on Windows it needs to be able to output a .dll.a along with the .dll).

Also do we need "Provide Unicode input over legacy API" when MT would be changed to support only the new unicode way?

nerzhul:
@sfan5 i totally agree. In fact we should discuss about a roadmap on this. We have multiple tries.

I think first we need to embed the vanilla 1.8.4, then implement cleanups, SDL, 1.9+opengles etc, one by one. But we need a clear roadmap first, we can use a board on this project with issues

PR #15, opened 2021-03-08
Remove trivial copy constructors and operators

sfan5:
Why? Compilers cannot optimize these away (even with `-O3`)!

<hr />

Consider this minimal example:

```cpp
#include <cstdint>

template <class T>
class vector3d
{
public:
	vector3d() : X(0), Y(0), Z(0) {}
#ifdef COPY
	vector3d(const vector3d<T>& other) : X(other.X), Y(other.Y), Z(other.Z) {}
	vector3d<T>& operator=(const vector3d<T>& other) { X = other.X; Y = other.Y; Z = other.Z; return *this; }
#endif
	T X, Y, Z;
};

extern vector3d<int32_t> g_vec;
vector3d<int32_t> func()
{
	vector3d<int32_t> m_vec = g_vec;
	return m_vec;
}
```

Compiled using `g++ test.cpp -c -S -o- -O3 -DCOPY`:
```asm
	movq	g_vec(%rip), %rdx
	movq	%rdi, %rax
	movq	%rdx, (%rdi)
	movl	8+g_vec(%rip), %edx
	movl	%edx, 8(%rdi)
	ret
```
`g++ test.cpp -c -S -o- -O3`:
```asm
	movl	8+g_vec(%rip), %edx
	movq	g_vec(%rip), %rax
	ret
```
`clang++ test.cpp -c -S -o- -O3 -DCOPY`:
```asm
	movq	%rdi, %rax
	movq	g_vec@GOTPCREL(%rip), %rcx
	movl	(%rcx), %edx
	movl	%edx, (%rdi)
	movl	4(%rcx), %edx
	movl	%edx, 4(%rdi)
	movl	8(%rcx), %ecx
	movl	%ecx, 8(%rdi)
	retq
```
`clang++ test.cpp -c -S -o- -O3`:
```asm
	movq	g_vec@GOTPCREL(%rip), %rcx
	movq	(%rcx), %rax
	movl	8(%rcx), %edx
	retq
```

It is most obvious in Clang's output:
With the copy constructor present, the compiler copies each vector component <b>individually</b>, with the constructor removed it can copy the underlying 12 bytes the most efficient way possible.
GCC is doing better, but still has to copy the vector twice since the vector is treated as a pointer. This is due to a different ABI, the vector can now be passed via registers.

<hr />

<b>Please squash when merging</b>

### DIFF
--- a/include/IAnimatedMeshMD3.h
+++ b/include/IAnimatedMeshMD3.h
@@ -156,12 +156,6 @@ namespace scene
 			position.X = 0.f;
 		}
 
-		// construct copy constructor
-		SMD3QuaternionTag( const SMD3QuaternionTag & copyMe )
-		{
-			*this = copyMe;
-		}
-
 		// construct for searching
 		SMD3QuaternionTag( const core::stringc& name )
 			: Name ( name ) {}
@@ -181,14 +175,6 @@ namespace scene
 			return Name == other.Name;
 		}
 
-		SMD3QuaternionTag & operator=( const SMD3QuaternionTag & copyMe )
-		{
-			Name = copyMe.Name;
-			position = copyMe.position;
-			rotation = copyMe.rotation;
-			return *this;
-		}
-
 		core::stringc Name;
 		core::vector3df position;
 		core::quaternion rotation;
@@ -200,12 +186,6 @@ namespace scene
 		SMD3QuaternionTagList()
 		{
 			Container.setAllocStrategy(core::ALLOC_STRATEGY_SAFE);
-		}
-
-		// construct copy constructor
-		SMD3QuaternionTagList(const SMD3QuaternionTagList& copyMe)
-		{
-			*this = copyMe;
 		}
 
 		virtual ~SMD3QuaternionTagList() {}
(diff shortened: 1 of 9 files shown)


nerzhul (approved):

Desour:
>With the copy constructor present, the compiler is forced to copy each vector component **individually**, with the constructor removed it can copy the underlying 12 bytes the most efficient way possible.

This, however, is not what your examples show.
gcc does copy X and Y at once, with and without `-DCOPY`.
However, the ABI seems to have changed. With `-DCOPY`, the signature of `func` is `vector3d<int32_t> *func(vector3d<int32_t> *ret_ptr)`; without `-DCOPY`, func is returning the vector in rax:rdx.
I guess c++ enforces this abi because because of the existence copy-constructor it might need to be able to have a pointer on the object.
(Also, I guess the return of the return pointer can be useful because it allows you to not save the return pointer.)

(The PR is good anyway.)

sfan5:
> gcc does copy X and Y at once, with and without -DCOPY.

You're correct, I didn't look closely at first and didn't bother to understand what gcc was doing.

> However, the ABI seems to have changed. With -DCOPY, the signature of func is vector3d<int32_t> *func(vector3d<int32_t> *ret_ptr); without -DCOPY, func is returning the vector in rax:rdx.
I guess c++ enforces this abi because because of the existence copy-constructor it might need to be able to have a pointer on the object.

This is pretty interesting. From a bit of ctrl+f in the [standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf) I'd say 3.9.2 is what enables the change (keyword is "<i>trivially copyable</i>"). This difference in ABI is not communicated via the symbol either, it's `_Z4funcv` in both cases.

PR #20, opened 2021-03-19
Readd irrUString.h

nerzhul:
Return back irrUString.h to its parent
Also discard all .so from git :)


v-rob:
Actually, core::ustring was never a part of Irrlicht, but was an extension that came with CGUITTFont.  Not that it matters, it's better in Irrlicht anyway...

PR #22, opened 2021-03-27
CMake Config and Version exports

JosiahWI:
All libraries using CMake should export their targets. This seems like a good time to add this and it will allow Minetest to include Irrlicht easier. I am not ready to merge this but I want input on what I need to change. I don't know how to properly provide the Config file so I've placed it in the home directory to be able to test that this works.


cmake/irrlichtmtConfig.cmake
```diff
@@ -0,0 +1,8 @@
+get_filename_component(IrrlichtMt_CMAKE_DIR "${CMAKE_CURRENT_LIST_DIR}" PATH)
+include(CMakeFindDependencyMacro)
+
+find_dependency(ZLIB REQUIRED)
```
sfan5, line 4: why is zlib here but not the rest?

is this needed at all?
looking at the generated `irrlichtmtTargets.cmake` there's this, which already specifies dependent libraries:
<pre>
set_target_properties(irrlichtmt::IrrlichtMt PROPERTIES
  INTERFACE_LINK_LIBRARIES "/usr/lib/x86_64-linux-gnu/libz.so;/usr/lib/x86_64-linux-gnu/libjpeg.so;/usr/lib/x86_64-linux-gnu/libpng.so;/usr/lib/x86_64-linux-gnu/libGL.so;/usr/lib/x86_64-linux-gnu/libX11.so;/usr/lib/x86_64-linux-gnu/libXxf86vm.so"
)
</pre>

sfan5:
> it will allow Minetest to include Irrlicht easier.

For the time being Minetest still needs to support building (the server) with upstream Irrlicht so we can't remove [`FindIrrlicht.cmake`](https://github.com/minetest/minetest/blob/master/cmake/Modules/FindIrrlicht.cmake) just yet.
This would simplify things for the client though as we don't have to search for the same libraries Irrlicht itself depends on.

I'm wondering two things though:
* Does this work with Windows DLLs? (I don't see why not but who knows)
* ~~I forgot what the second was~~

cmake/irrlichtmtConfig.cmake
JosiahWI: Thanks for bringing that to my attention. That was in an example I used as a reference, so I added one to remind me to research it and figure out whether it's needed. The explanation wasn't very clear.

JosiahWI:
It should probably be tested with DLLs, unfortunately I don't have a Windows environment to test it in myself. I didn't change the destination of the library and include install locations, so it should work, but as you said who knows.

cmake/irrlichtmtConfig.cmake
JosiahWI: Ah. It looks like those are needed if the Targets file references other targets. In Irrlicht's case, it's unecessary, because Irrlicht doesn't link zlib etc. as targets (although it should). I'll remove the dependency for now, and then I can re-add it in a separate PR to link those libraries as targets.

JosiahWI:
Reading the CMake find_package documentation, I see that there are a number of places CMake will automatically look for Config files. We can also export a file declaring Irrlicht's versioning information, which CMake can use to enforce version requirements when finding the package, but do we want that?

JosiahWI:
I'm ready to merge this if it's accepted. I've tested it on Debian and it each part worked as expected.

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.5)
 
-project(Irrlicht CXX)
+project(Irrlicht
+	VERSION 1.9.0
```
sfan5, line 4: One more thing: should should be populated from `${VERSION}`
(just move a `set_property` or whatever call down below that)

nerzhul:
> > it will allow Minetest to include Irrlicht easier.
> 
> For the time being Minetest still needs to support building (the server) with upstream Irrlicht so we can't remove [`FindIrrlicht.cmake`](https://github.com/minetest/minetest/blob/master/cmake/Modules/FindIrrlicht.cmake) just yet.
> This would simplify things for the client though as we don't have to search for the same libraries Irrlicht itself depends on.
> 
> I'm wondering two things though:
> 
>     * Does this work with Windows DLLs? (I don't see why not but who knows)
> 
>     * ~I forgot what the second was~

why do we keep this compatiblity ?

CMakeLists.txt
JosiahWI: I'd like to propose another solution to this. VERSION was introduced to do exactly what ${VERSION} is doing now. I'd like to remove ${VERSION} entirely and use the provided system which is supported since version 3.0, if this is ok with you.

CMakeLists.txt
```diff
@@ -9,8 +9,7 @@ set(IRRLICHT_VERSION_MAJOR 1)
 set(IRRLICHT_VERSION_MINOR 9)
 set(IRRLICHT_VERSION_RELEASE 0)
```
sfan5, line 10: Remove these too, then?

sfan5:
> why do we keep this compatiblity ?

It turns out we need to keep it anyway to find the Irrlicht headers for the server (without it even being built).

JosiahWI:
> > why do we keep this compatiblity ?
> 
> It turns out we need to keep it anyway to find the Irrlicht headers for the server (without it even being built).

The IrrlichtMt headers don't work?

sfan5:
I used Irrlicht and IrrlichtMt interchangeably here. Point is we can't require a `IrrlichtMtConfig.cmake` for server builds.

JosiahWI:
Aha I wasn't thinking through this carefully, you are right and I understand the issue now. This can easily be fixed, but I'll do it in a separate PR.

JosiahWI:
Hopefully this is good to merge now. And I want to thank you guys for the opportunity to work on this, I don't want to disrespect the work that others have put into setting this up.

JosiahWI:
Don't merge yet, there's a bug (in the current branch too) that will break irrlicht's install if configured as a submodule.

JosiahWI:
The bug is fixed, it should be good now.

PR #24, opened 2021-03-31
Expose CGUIEditBox as an API

yw05:
This PR exposes `source/CGUIEditBox.h` and makes it support features used by `GUIEditBox` from Minetest.
# Changes
- Remove `-fno-rtti` from the C++ compiler flags
# To-Do
- [x] Implement the `writable` field of Minetest `GUIEditBox`
- [ ] Implement vertical scrollbar
- [x] Bug: It is possible to look at the content of password fields with the overwrite mode enabled
- [x] (Done locally) Make Minetest `GUIEditBox` inherit `CGUIEditBox` instead of `IGUIEditBox`
- [ ] Clean up `GUIEditBox` and `GUIEditBoxWithScrollBar` code form the Minetest source (a separate PR will be made)


sfan5:
Hmm I don't know, why not add the necessary functions to `IGUIEditBox` and keep the implementation hidden?

yw05:
My original idea is that `CGUIEditBox` already contains part of the implementation, so I can just port some of the code from Minetest to make it complete (or should I just `cp minetest/src/gui/guiEditBox.cpp irrlicht/source/Irrlicht/CGUIEditBox.cpp`.

yw05:
I think I misunderstood the way the `C*` classes work after some work on IME support. Closing.

nerzhul:
i have the same idea as you @yw05 but it's harder than expected :)
but yeah at the end i'd like to see irrlicht engine owning the whole GUI global logic from MT

yw05:
IMO that would be a good thing, but I originally thought that the `CGUI*` classes where not used at all until I ran GDB on minetest and realized I was wrong.

PR #29, opened 2021-04-19
Fix Minetest chatroom keyboard acts strange

Emojigit:
Close minetest/minetest#11190


yw05:
There's an unrelated bug in Minetest where MT segfaults when restarting the Fcitx5 or IBus. I posted the patch to fix minetest/minetest#11190 (it is also on yw05/irrlicht@aeef1b76e), but I would like to fix that segfault bug as well before filing the PR.

Emojigit:
Replaced by https://github.com/minetest/irrlicht/commit/21cb990e75843d62182e82618a1e4ed7e1a9ecdf

PR #32, opened 2021-04-23
Work around bug installing targets created in a subdirectory

### LINKED ISSUE #28: CMake install target breaks build on CMake 3.7
Maybe wrong name?
```
CMake Error at CMakeLists.txt:32 (install):
  install TARGETS given target "IrrlichtMt" which does not exist in this directory.
-- Configuring incomplete, errors occurred!
```

JosiahWI:
The objective of this PR is to fix #28 . We need this tested by someone who has a CMake version less than 3.13 to see whether the changes are sufficient.

### DIFF
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -408,8 +408,6 @@ else()
 	set(INSTALL_TARGETS IrrlichtMt)
 endif()
 
-# Installation of library and headers.
-include(GNUInstallDirs)
 install(TARGETS ${INSTALL_TARGETS}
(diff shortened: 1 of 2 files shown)


source/Irrlicht/CMakeLists.txt
```diff
@@ -403,7 +403,12 @@ if(WIN32)
 endif()
 
 if(ANDROID)
```
sfan5, line 405:
```suggestion
# Installation of library

if(ANDROID)
```

sfan5 (approved):
Works on Debian 9 w/ CMake 3.7.2

PR #33, opened 2021-04-27
cleanup C++0x support in irrUString.h

nerzhul:


include/irrUString.h
```diff
@@ -83,17 +72,10 @@ static const irr::u16 UTF16_LO_SURROGATE = 0xDC00;
 
 namespace irr
 {
-
 	// Define our character types.
-#ifdef USTRING_CPP0X_NEWLITERALS	// C++0x
```
sfan5, line 88: other way around, `char32_t` is a C++11 type so we should use it

nerzhul: @sfan5 you mean we should keep the mapping here ?

include/irrUString.h
```diff
@@ -886,10 +853,7 @@ class ustring16
 		for (; first != last; ++first)
 			append((uchar32_t)*first);
 	}
-#endif
 
-
-#ifndef USTRING_CPP0X_NEWLITERALS
```
sfan5, line 892: If you remove <i>exactly</i> the function inside `#ifndef USTRING_CPP0X_NEWLITERALS` then it should work

include/irrUString.h
```diff
@@ -2091,28 +2019,6 @@ class ustring16
 	}
 
 
-#ifdef USTRING_CPP0X_NEWLITERALS
```
sfan5, line 2094: There's also a few more blocks of code that should stay.

Issue #34, opened 2021-05-01
Add verbose flag in CI build


JosiahWI:
This is a feature request to improve the process of debugging the build system. It's a very easy change, but I'm opening an issue because I can see the counterargument that the build will seem cluttered. Would this be an acceptable change to PR?

sfan5:
You want to change `make` to `make VERBOSE=1` or is there something else?

JosiahWI:
I'm not familiar with make options. I want to do the equivalent of `cmake --build <build_dir> -v`

sfan5:
That would be the equivalent, yes. Sounds fine to me.
(would probably suffice to do this for one of the Linux builds though)

PR #35, opened 2021-05-01
Add verbose flag to CI build

JosiahWI:
Make is now called with VERBOSITY=1 in the linux GL build in the CI, and in the mingw script which is run by both of the Windows builds in the CI.


JosiahWI:
This will close #34.

PR #36, opened 2021-05-18
Use relative include dir to fix subdir usage

NeroBurner:
To make it possible to use IrrlichtMt with `add_subdirectory()` command
use a relative include path instead of using `CMAKE_SOURCE_DIR`. When
using IrrlichtMt as a subdirectory the `CMAKE_SOURCE_DIR` variable points
to the root of the Minetest repo instead of the root of the IrrlichtMt
source directory, which breaks the build because the `IrrCompileConfig.h`
include can't be found.

### DIFF
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -1,5 +1,5 @@
 include_directories(
-	${CMAKE_SOURCE_DIR}/include
+	${CMAKE_CURRENT_SOURCE_DIR}/../../include
 	${CMAKE_CURRENT_SOURCE_DIR}
 )
 add_definitions(-DIRRLICHT_EXPORTS=1)


NeroBurner:
@nerzhul @sfan5 please review 🙇‍♂️

sfan5:
#31 has been on my TODO list longer and I _think_ it makes these changes unnecessary/redundant, which is why I haven't yet.

JosiahWI:
This should still be done to keep things safe in case someone adds a subdir, and it's nice to keep that compatibility (that's why I made an alias target in the aforementioned PR). I've also fixed some paths similar to this one that should've been based in the project root, but weren't. Things could mysteriously break for people if they try to add this as a subdirectory and the paths are based in the source dir.

JosiahWI:
I tested and the current version works as a subdirectory already on Linux, because the target include directories have the correct paths, but this should still be added.

PR #39, opened 2021-06-18
Set includes and libs on object targets

NeroBurner:
Instead of using the global `include_directories()` and
`link_libraries()` calls use the target versions
`target_include_directories()` and `target_include_libraries()` with
`PRIVATE` keyword instead.

### DIFF
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -65,7 +65,7 @@ else()
 endif()
 
 # Includes for object libraries.
-include_directories(
+set(link_includes
 	"${PROJECT_SOURCE_DIR}/include"
 	"${CMAKE_CURRENT_SOURCE_DIR}"
 
@@ -100,7 +100,6 @@ set(link_libs
 	${X11_X11_LIB}
 	${X11_Xxf86vm_LIB}
 )
-link_libraries(${link_libs})
 
 # Source files
 
@@ -401,18 +400,22 @@ add_library(IRRGUIOBJ OBJECT
 
 option(BUILD_SHARED_LIBS "Build shared library" TRUE)
 
-# IrrlichtMt target
-add_library(IrrlichtMt
-	$<TARGET_OBJECTS:IRRMESHOBJ>
-	$<TARGET_OBJECTS:IRROBJ>
-	$<TARGET_OBJECTS:IRRPARTICLEOBJ>
-	$<TARGET_OBJECTS:IRRANIMOBJ>
-	$<TARGET_OBJECTS:IRRVIDEOOBJ>
-	$<TARGET_OBJECTS:IRRSWRENDEROBJ>
-	$<TARGET_OBJECTS:IRRIOOBJ>
-	$<TARGET_OBJECTS:IRROTHEROBJ>
-	$<TARGET_OBJECTS:IRRGUIOBJ>
-)
+add_library(IrrlichtMt)
+foreach(object_lib
+	IRRMESHOBJ
+	IRROBJ
+	IRRPARTICLEOBJ
+	IRRANIMOBJ
+	IRRVIDEOOBJ
+	IRRSWRENDEROBJ
+	IRRIOOBJ
+	IRROTHEROBJ
+	IRRGUIOBJ)
+	# Set include directories for object library compilation
+	target_include_directories(${object_lib}     PRIVATE ${link_includes})
+	# Add objects from  object library to main library
+	target_sources(IrrlichtMt PRIVATE $<TARGET_OBJECTS:${object_lib}>)
+endforeach()
 
 # Alias target provides add_submodule compatibility
 add_library(IrrlichtMt::IrrlichtMt ALIAS IrrlichtMt)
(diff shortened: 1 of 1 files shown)


source/Irrlicht/CMakeLists.txt
```diff
@@ -396,6 +395,24 @@ add_library(IRRGUIOBJ OBJECT
...
+target_include_directories(IRRGUIOBJ      PRIVATE ${link_includes})
+target_link_libraries(IRRMESHOBJ     PRIVATE ${link_libs})
+target_link_libraries(IRROBJ         PRIVATE ${link_libs})
+target_link_libraries(IRRPARTICLEOBJ PRIVATE ${link_libs})
+target_link_libraries(IRRANIMOBJ     PRIVATE ${link_libs})
+target_link_libraries(IRRVIDEOOBJ    PRIVATE ${link_libs})
+target_link_libraries(IRRSWRENDEROBJ PRIVATE ${link_libs})
+target_link_libraries(IRRIOOBJ       PRIVATE ${link_libs})
+target_link_libraries(IRROTHEROBJ    PRIVATE ${link_libs})
+target_link_libraries(IRRGUIOBJ      PRIVATE ${link_libs})
```
sfan5, line 415: why are all these needed, doesn't the target already get the includes and libs?

NeroBurner: the IrrlichtMt target yes, but those are object-libraries, and they need the information on how they get compiled as well.

One alternative: we could change the CMakeLists some more to add the files to the `IrrlichtMt` target instead of creating object-libraries. Then we would need only one `target_link_libraries()` and `target_include_directories()` call. This would be a greater change (I think), I just wanted a small change.

I assume the object-libraries are used here to have a logical grouping of files, but that's just my guess

NeroBurner: apple has got a special handling with one of the object-libraries called `IRROTHEROBJ`. I can keep just this one object-library and add the remaining files directly to the `IrrlichtMt` library target

```cmake
elseif(APPLE)
	# Build all IRROTHEROBJ sources as objc++, including the .cpp's
	set_target_properties(IRROTHEROBJ PROPERTIES COMPILE_OPTIONS "-xobjective-c++")
	target_sources(IRROTHEROBJ PRIVATE
		CIrrDeviceOSX.mm
		CNSOGLManager.mm
	)
```

JosiahWI: I also fixed this today just now. I've been checking my email, but this PR was so recent I missed it. I don't know if NeroBurner is in the IRC channel; I was discussing the issue and my progress on it there.

JosiahWI: The change you made is effectively the exact same change I made, and the secondary solution you proposed was also my plan B, so I think we should keep this PR and roll with it.

JosiahWI:
I also think I should mention for completeness for the people who aren't familiar with how this works, when IrrlichtMt is built as a static lib, CMake is smart enough to tell Minetest it needs to link the libraries even though they're PRIVATE dependencies. If IrrlichtMt is built as a dynamic library, Minetest will not be told anything about Irrlicht's dependencies.

source/Irrlicht/CMakeLists.txt
sfan5: There are ways to avoid this big list.
You can write a macro that wraps `add_library(IRRWHATEVEROBJ OBJECT [...])` and automatically calls `target_include_directories` + `target_link_libraries`

> I assume the object-libraries are used here to have a logical grouping of files, but that's just my guess

Correct, this logical grouping was taken from the old Irrlicht Makefile. I don't think that should be dropped.

NeroBurner: I finally understood your remark about the useless `target_link_libraries()` calls and hence removed them with the following explanation

```
 Don't link dependencies to object libraries

The dependencies are libraries only, the headers are added to the build
instructions in a separate call. So the `target_link_libraries()` calls
are not needed for the object-libraries to be built.
```

JosiahWI: That's the way I did it in the first place and it broke the build on MinGW or something I thought. Maybe it was something else that I missed.

NeroBurner:
@sfan5 You said to keep the logical grouping

> Correct, this logical grouping was taken from the old Irrlicht Makefile. I don't think that should be dropped

I've consolidated the object libraries (except `IRROTHEROBJ` which has special handling under Android and Apple). The logical grouping is now in the grouping of the source files and the comments between the source file blocks. This minimizes the needed include and link instructions. If this is an unwanted change I'll remove/revert the commit again

https://github.com/minetest/irrlicht/pull/39/commits/9645091d486d49da2ddba7cb46e31787958837a9

source/Irrlicht/CMakeLists.txt
NeroBurner: It seems to build with https://github.com/NeroBurner/minetest/actions/runs/958376062, but that's using https://github.com/NeroBurner/minetest/tree/use_builtin_irrlichtmt_with_ci branch

screenshot of the GitHub Actions Pipeline with successful Mingw (only MSVC fails as expected)

image

sfan5:
> You can write a macro that wraps `add_library(IRRWHATEVEROBJ OBJECT [...])` and automatically calls `target_include_directories` + `target_link_libraries`

try this one? It would be cleaner while still keeping the different targets.

NeroBurner:
@sfan5 **TL;DR** I decided to go the `foreach()` route to satisfy the "don't repeat yourself" principle

The long version:

The `macro()` or `function()` route would have been feasible, but I decided to go a different route, because I couldn't think of a good way to have "the first argument" and then "the remaining arguments". Otherwise I would have had to call [cmake_parse_arguments](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html) which would have resulted in calls like the following (in case of `function`)

```cmake
add_object_library_and_include_target(TARGET IRRGUIOBJ
	SOURCES
	CGUIButton.cpp
	CGUICheckBox.cpp
	...
	CGUIProfiler.cpp
	INCLUDES ${link_includes}
)
```

with `macro()` the last `INCLUDES` line wouldn't be needed, but I don't like the variable spilling nature of `macro()`. With argument passing `macro()` has some pitfalls ( [see documentation 'Argument Caveats' for more info](https://cmake.org/cmake/help/latest/command/macro.html?highlight=macro#argument-caveats) ).

I also don't like the non-standard feel of the custom `function()`/`macro()` solution. So I went for the simple and [boring solution](https://about.gitlab.com/handbook/values/#boring-solutions): `foreach()`.

It is standard-CMake which makes it easier understandable for me (and hopefully others as well). The line-count for this solution is low as well, which is good for readability and again easier to understand

---

If the solution is acceptable I'll sqash-rebase if you want

NeroBurner:
Squashed

sfan5 (approved):

PR #41, opened 2021-06-21
Fix `getViewPort` returning incorrect results

v-rob:
`IVideoDriver::getViewPort` has a bug where it only returns viewports set explicitly by `setViewPort`.  However, `OnResize` and `setRenderTarget(Ex)` also set the viewport since the drawing area has changed, but they only set it for OpenGL, not for `getViewPort`.  This PR fixes that.

Minetest currently doesn't do enough to run into any problems with this, but https://github.com/minetest/minetest/pull/10801 does, and this PR makes it work without a hacky workaround.

### DIFF
--- a/source/Irrlicht/COGLES2Driver.cpp
+++ b/source/Irrlicht/COGLES2Driver.cpp
@@ -2101,9 +2101,7 @@ COGLES2Driver::~COGLES2Driver()
 
 	void COGLES2Driver::setViewPortRaw(u32 width, u32 height)
 	{
-		if (width > 0 && height > 0)
-			CacheHandler->setViewport(0, 0, width, height);
-
+		CacheHandler->setViewport(0, 0, width, height);
 		ViewPort = core::recti(0, 0, width, height);
 	}
 
@@ -2261,7 +2259,7 @@ COGLES2Driver::~COGLES2Driver()
 	void COGLES2Driver::OnResize(const core::dimension2d<u32>& size)
 	{
 		CNullDriver::OnResize(size);
-		setViewPort(core::recti(0, 0, size.Width, size.Height));
+		CacheHandler->setViewport(0, 0, size.Width, size.Height);
 		Transformation3DChanged = true;
 	}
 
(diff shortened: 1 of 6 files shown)


source/Irrlicht/COGLES2Driver.cpp
```diff
@@ -2099,6 +2099,15 @@ COGLES2Driver::~COGLES2Driver()
 	}
 
 
+	void COGLES2Driver::setViewPortRaw(const core::rect<s32>& vp)
```
sfan5, line 2102: This should be a `core::dimension2d<u32>`, I don't see this being called with negative values anywhere

v-rob: (For that matter, `setViewPort` should be a core::rect<u32>, but who said Irrlicht made sense?)

Fixed.

source/Irrlicht/COGLES2Driver.cpp
```diff
@@ -2099,6 +2099,15 @@ COGLES2Driver::~COGLES2Driver()
 	}
 
 
+	void COGLES2Driver::setViewPortRaw(u32 width, u32 height)
+	{
+		if (width > 0 && height > 0)
```
sfan5, line 2104: redundant, these are unsigned

source/Irrlicht/COGLES2Driver.cpp
```diff
@@ -2252,7 +2261,7 @@ COGLES2Driver::~COGLES2Driver()
 	void COGLES2Driver::OnResize(const core::dimension2d<u32>& size)
 	{
 		CNullDriver::OnResize(size);
-		CacheHandler->setViewport(0, 0, size.Width, size.Height);
+		setViewPort(core::recti(0, 0, size.Width, size.Height));
```
sfan5, line 2264: changing `CNullDriver::OnResize(` to set `ViewPort` in all cases appears to work just fine, then you don't need to change the implementation in each driver

PR #47, opened 2021-07-14
Add macOS build

FnControlOption:
https://github.com/FnControlOption/irrlicht/actions/runs/1029348615#artifacts

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,15 +96,5 @@ jobs:
 
       - name: Build
         run: |
-          cmake .
+          cmake . -DCMAKE_FIND_FRAMEWORK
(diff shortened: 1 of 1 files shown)


sfan5:
Looks good but I don't think we need it to produce an artifact, or do you have a plan for that?

FnControlOption:
Nope, I'll remove it. Also, I'll add `-DCMAKE_FIND_FRAMEWORK=LAST` so that the build doesn't use `libpng` and `libjpeg` from `Mono.framework`. These versions are outdated:

```
-- Found JPEG: /usr/local/lib/libjpeg.dylib (found version "80") 
-- Found PNG: /usr/local/lib/libpng.dylib (found version "1.4.12") 
```

PR #48, opened 2021-07-22
Delete unused features

hecktest:
This PR deletes features that we do not use:
- Non-GL renderers
- Terrain
- Particles
- Selectors
- Octrees
- Animators (rotators, spline followers, etc.)
- Primitive scene nodes
- Text scene nodes
- Light nodes (SLight still exists... for now)
- Stencil shadows
- Collision manager
- Light manager
- Model texture loaders
- Model writers
- Redundant file formats
- Dead file formats
- Quake and other engines interop

About 50% of the engine has been removed (including already disabled features), bringing the code to roughly the size of our own.
This greatly reduces the compile time, output binary size, improves the readability of some parts (like the render loop), and reduces the cognitive load and maintenance cost of any further modifications to the engine.

File formats have been triaged rather aggressively - I've left only PNG, JPEG, X, B3D, OBJ.
BMP is also enabled because the null video driver would not compile without it.

Despite our weak endorsement of MD2 and the curiosity of having a format with vertex animation support, I decided to axe it because:
- You'll have a hard time finding a working md2 exporter for any 3D program worth a damn
- Vertex animation is inefficient and probably a bad idea to send over the network
- You can't perform any bone attachment or bone positioning with no bones to speak of.

Thus I don't think that anybody actually uses it, and our API is biased towards skinned meshes anyway.

In order to test, you will have to re-link your copy of Minetest after compiling this.


sfan5:
Note that the "linux-gles" CI run builds all examples, remove whatever you think should go. The tiny automated test could render a static mesh instead I guess.

Also for merging later it'd be good if you could reauthor the commits so that Github can automatically match them to an account.

Calinou:
> Despite our weak endorsement of MD2 and the curiosity of having a format with vertex animation support, I decided to axe it because:
>
> - You'll have a hard time finding a working md2 exporter for any 3D program worth a damn
> - Vertex animation is inefficient and probably a bad idea to send over the network
> - You can't perform any bone attachment or bone positioning with no bones to speak of.

I'm not aware of any Minetest mod using a MD2 model, so this is probably fine.

hecktest:
You mean the example demos? I think all of these should go.
Likewise if any unit test breaks I would just delete it, because they'll keep breaking as we shuffle things around.
It would be more valuable to build MT itself as part of the CI job, guess that's one more reason to merge the repos one day.

rollerozxa:
With this PR, stripped libIrrlichtMt.so went down from 4.7MB to 2.7MB. Very nice.

I compiled it, tested opening MInetest and playing various games (briefly hopping into Development Test, Minetest Game, MineClone2 and NodeCore) and it all seems to work like it should.

sfan5:
> Likewise if any unit test breaks I would just delete it, because they'll keep breaking as we shuffle things around.

It's not really unit tests, just something super simple to assert that it doesn't crash on startup and other trivial things.

hecktest:
Well, it sure needs a replacement for that .md2 model it can't use.
I think I have an idea....
cool hat

hecktest:
Do we use the /tests/ folder in CI at all?

hecktest:
Looks like we don't, I'd rather keep those deleted due to reasons mentioned above. (most of these tests are now broken and will keep breaking)

sfan5 (approved):

AFCMS:
Is tga image format still supported???

AFCMS:
mcl2 use it

AFCMS:
removing tga support is the direct way to https://www.reddit.com/r/MineClone2/comments/oepcc3/what_should_a_minetest_engine_fork_for_mineclone2/
so plz if u removed tga support, revert it :pray:

rollerozxa:
TGA image loading seems to have been removed.

AFCMS:
noooo :sob:

sfan5:
> mcl2 use it and removing support will break mcl2 AND servers
https://git.minetest.land/MineClone2/MineClone2/src/branch/master/mods/ITEMS/mcl_maps

You appear to already have a TGA encoder library that writes uncompressed files, you can easily replace this with code that writes a BMP.

wsor4035:
re tga support being removed(undocumented, not supported), see this xkcd: https://xkcd.com/1172/

AFCMS:
> > mcl2 use it and removing support will break mcl2 AND servers
> > https://git.minetest.land/MineClone2/MineClone2/src/branch/master/mods/ITEMS/mcl_maps
> 
> You appear to already have a TGA encoder library that writes uncompressed files, you can easily replace this with code that writes a BMP (it's a bit simpler in fact).

oh
Any mt mod already using this? So this would only break servers?

sfan5:
> So this would only break servers?

No, you're correct that this would affect any servers running MCL with 5.5.0-dev clients.
Like I said it would be fairly simple to edit [this implementation](https://git.minetest.land/MineClone2/MineClone2/src/branch/master/mods/CORE/tga_encoder/init.lua) to generate BMP images instead.

hecktest:
In the long term we should just add a PNG writer API to officially support init-time image generation without hacks like this.

AFCMS:
> In the long term we should just add a PNG writer API to officially support init-time image generation without hacks like this.

100% aggree

hecktest:
Here it is.

minetest/minetest#11485

nerzhul:
i know we merged this @hecktest but why remove the collision manager ? it's not interesting in our context ??

hecktest:
It really isn't, and it's not about what's "interesting" as opposed to what we actually are using.
Irrlicht code is usually lousy enough that it's not worth keeping around just in case we might decide to use it.
Irrlicht is also a client only dependency so we could not use any of this on the server.

HybridDog:
Would it make sense to decode image and sound files with libavcodec (ffmpeg)?
ffmpeg supports many image and audio formats including webp and opus.

Calinou:
> Would it make sense to decode image and sound files with libavcodec (ffmpeg)?
> ffmpeg supports many image and audio formats including webp and opus.

ffmpeg is a large library with patent issues depending on the features you enabled. It's also not meant to be used for decoding images in a game engine (although it's fine for videos).

v-rob:
If we wanted more audio format support or support of videos, then ffmpeg might be good.  Otherwise, we're probably fine how we are.

appgurueu:
Please do not axe MD2. It is a simple file format which can comfortably be used for generating dynamic models. Same goes for TIFF; Mineclone already uses that (they should switch to PNG now that an exporter is available though). Note that axing file formats breaks compat and would have to wait for 6.x.

hecktest:
@appgurueu 
Since when is MD2 simpler or more comfortable for procgen than .obj or .x?

MD2 has the following problems:
- Binary (not that comfortable for procgen unless you use FFI)
- No living exporters for common tools like Blender
- Not robust or standardized; the only documentation is the Q2 source code and things people have reverse engineered, it's just id's inhouse model format for q2
- set/get bone transform does not work as MD2 uses vertex animation
- Vertex animation is orders of magnitude less efficient than skeletal, file sizes would be terrible
- Vertex animation of this kind cannot easily be hardware accelerated - if we switch to hardware skinning, MD2 would be a performance bomb for unwary users
- Each additional file format that the client can interpret is extra attack surface for malicious servers
- More code for us to maintain as part of IrrMT
- No documented examples of anyone using it in Minetest; thus there's no excuse to keep this dead weight around

Since our skeletal animation does support "difficult" things like scale keyframes, most of the things that you'd want vertex animation for are probably achievable with bones at a lower cost. We already support .b3d as an efficient binary format which is probably no harder to emit than one of id's internal formats.

PR #49, opened 2021-07-28
Export targets to build tree

JosiahWI:
It is insecure and inconvenient to necessitate installing IrrlichtMt to access the target export. A copy of the export is now placed in the build tree during configuration so that it can be directly accessed from any project by pointing find_package() to look at the IrrlichtMt source directory.


sfan5 (approved):
import from build tree works
standalone install still works

PR #52, opened 2021-08-05
Add a unified cross platform OpenGL core profile binding

hecktest:
This PR adds an OpenGL wrangler code generator script that generates a shared core profile binding for all supported platforms. Only core and ARB functions from `glcorearb.h` are provided, and a minimum feature level of OpenGL ES 2.0 is assumed. Other headers are searched for compatible procedure aliases that may be used as a fallback.

This allows us to call OpenGL directly from the Minetest source. Just include `<mt_opengl.h>` and you may access a global `GL` struct containing all the procedure pointers with un-mangled names, for example:
```
GL.DrawArrays
GL.UseProgram
GL.Enable
GL.GenBuffers
```
as well as necessary constants, such as
```
GL.ALPHA_TEST
GL.FRONT
GL.DST_COLOR
GL.TEXTURE0
```

The pointers are valid as long as your OpenGL driver contains the relevant function, a GL IVideoDriver has been created at least once, and you have an appropriate GL version or an extension that provides the function.

iOS was not considered in this PR as Apple does not document how to load GL procedures on iOS or whether it's supported at all. Anyone who cares about compiling Minetest for that platform is free to provide the implementation of `IContextManager::getProcAddress` for it.

### How to re-generate the loader

**You shouldn't have to do this**, artifacts are committed. But if you want to, you will need the following:
- A standalone Lua 5.1 compatible runtime
- GL core header: `glcorearb.h`
- GLES2 extension header: `gl2ext.h`

Headers can be obtained here: https://github.com/KhronosGroup/OpenGL-Registry/tree/master/api

Then, you need to run `scripts/BindingGenerator.lua`, read the script for instructions.

### How to test

To test this, try compiling and running the `gl-test` branch in my MT repo:
https://github.com/hecktest/minetest/tree/gl-test
It should spit out a few GL function pointers to `warningstream`.
`glEnable` and `glUseProgram` should never be null, others could be null depending on GL version, extension presence and platform.

### To do
- [x] Confirm it works on Windows
- [x] Confirm it works on Linux et al.
- [x] ~~Confirm it works on OSX~~ _OSX building is broken at the moment._
- [x] Make the GLES2 driver initialize the pointers
- [x] ~~Confirm it works on Android~~ _Good luck finding someone able and willing. The CI test for EGL/GLES does pass._
- [x] Consider whether the fallbacks are a good idea _(Yes to aliasing ARB - OES, no to everything else)_
- [x] Add an extension checker
- [x] Trim some fat from the generated headers (do we really need all these _NV defines?)
- [x] Convert numerical defines into consts
- [x] ~~Do something to support the null device~~ _Deferred_

## This PR is ready for review.

The following tasks have been deferred:
- Android testing (lack of manpower; I tacitly assume it will work since Android is just Linux with GLES
- OSX testing (Broken build system, needs updating with proper C++11 support)
- Null device handling (CI may temporarily use Mesa; calling GL functions with the Null Video Driver is currently an instant crash)


JosiahWI:
It works on Debian Bullseye (testing).

hecktest:
@JosiahWI 
You mentioned something in the IRC about PRing the c++11 requirement in the irrlicht repo; you should know that just setting `std=c++11` and disabling strict ANSI (as irrlicht demands) ended up breaking other builds **in this particular PR** and I had to revert it. **If this were to happen in your isolated PR, we would not be able to catch this issue** so if you want to tackle the OSX build problem, please test your PR in a separate branch with this merged on top.

include/IContextManager.h
```diff
@@ -48,6 +49,9 @@ namespace video
 			Note that only 1 thread at a time may access an OpenGL context.	*/
 		virtual bool activateContext(const SExposedVideoData& videoData, bool restorePrimaryOnZero=false) =0;
 
+		//! Get the address of any OpenGL procedure (including core procedures).
+		virtual void* getProcAddress(const std::string procName) =0;
```
sfan5, line 53:
```suggestion
		virtual void* getProcAddress(const std::string &procName) =0;

```

scripts/BindingGenerator.lua
```diff
@@ -0,0 +1,446 @@
...
+]];
+f:write( pointers:Concat( "\n" ) );
+f:write( "\n\n" );
+f:write( cppConsts:Concat( "\n" ) );
+f:write( "\n\n" );
+f:write( "};\n" );
+f:write( "\n//Global GL procedures object.\n" );
+f:write( [[
+#ifdef WIN32
+	#ifdef MINETEST_CLIENT
```
sfan5, line 406: I think it'd be better to invert this, less error prone. We can have CMake set the define for us while IrrMt is built

scripts/BindingGenerator.lua
```diff
@@ -0,0 +1,446 @@
...
+
+]];
+f:write( pointers:Concat( "\n" ) );
+f:write( "\n\n" );
+f:write( cppConsts:Concat( "\n" ) );
+f:write( "\n\n" );
+f:write( "};\n" );
+f:write( "\n//Global GL procedures object.\n" );
+f:write( [[
+#ifdef WIN32
```
sfan5, line 405: (used everywhere else too)
```suggestion
#ifdef _WIN32
```

scripts/BindingGenerator.lua
```diff
@@ -0,0 +1,446 @@
...
+	#ifdef MINETEST_CLIENT
+		#define DLL_SHARED __declspec(dllimport)
+	#else
+		#define DLL_SHARED __declspec(dllexport)
+	#endif
+#else
+	// On Linux, this supposedly works out of the box.
+	#define DLL_SHARED
+	// And on OSX, everything is linked statically anyway.
```
sfan5, line 414:
```suggestion
	// Other platforms don't need this
	#define DLL_SHARED
```

scripts/BindingGenerator.lua
```diff
@@ -0,0 +1,446 @@
...
+
+]];
+f:write( loader:Concat() );
+f:write[[
+
+	// get the extension string, chop it up
+	std::string ext_string = std::string((char*)GetString(EXTENSIONS));
+	std::stringstream ext_ss(ext_string);
+	std::string tmp;
+	while (getline(ext_ss, tmp, ' '))
```
sfan5, line 441: I don't see an `use namespace` so how does this even work?
```suggestion
	while (std::getline(ext_ss, tmp, ' '))
```

source/Irrlicht/CEGLManager.cpp
```diff
@@ -590,6 +594,19 @@ const SExposedVideoData& CEGLManager::getContext() const
 	return Data;
 }
 
+void* CEGLManager::getProcAddress(const std::string name)
+{
+	void* proc = NULL;
+	proc = (void*)eglGetProcAddress(name.c_str());
+	if (!proc) { // fallback
```
sfan5, line 601: Do we actually know if this is needed or is it here just in case?

source/Irrlicht/CNSOGLManager.mm
```diff
@@ -225,11 +226,27 @@
...
+	NSSymbol symbol = NULL;
+	// Allocate a buffer for the name, an underscore prefix, and a cstring terminator.
+    char *mangledName = (char*)malloc(procName.size() + 2);
+    strcpy(mangledName + 1, procName.c_str());
+    mangledName[0] = '_';
+    if (NSIsSymbolNameDefined(mangledName))
+        symbol = NSLookupAndBindSymbol(mangledName);
+    free(mangledName);
+
+    return symbol ? NSAddressOfSymbol(symbol) : NULL;
```
sfan5, line 247:
```suggestion
	std::string mangledName = "_" + procName;
	if (NSIsSymbolNameDefined(mangledName.c_str()))
		symbol = NSLookupAndBindSymbol(mangledName.c_str());
	return symbol ? NSAddressOfSymbol(symbol) : NULL;
```

source/Irrlicht/CEGLManager.cpp
hecktest: SDL does it so I do it as well, it doesn't hurt in any case.

scripts/BindingGenerator.lua
hecktest: I have no idea, but it doesn't hurt to add.

sfan5 (approved):

PR #58, opened 2021-08-21
Require C++11 standard to fix issue on OSX builds

JosiahWI:


JosiahWI:
This was not supposed to pass the checks, according to @hecktest. I have an uneasy feeling that I've overlooked something.

sfan5:
You overlooked that we already do this https://github.com/minetest/irrlicht/blob/4ba0f8d1009275899dd9d244f92527274e93e65b/source/Irrlicht/CMakeLists.txt#L13-L14

JosiahWI:
Well, that's quite the oversight. Does this mean the OSX issue is fixed? This PR can be closed then.

sfan5:
Yes.

PR #63, opened 2021-08-27
Add support for OpenBSD

### LINKED ISSUE #37: Request: Add support for OpenBSD

JosiahWI:
I can confirm this compiles Irrlicht on OpenBSD 6.9. I will make sure Minetest compiles shortly.

Resolves #37.


JosiahWI:
Minetest is confirmed to compile on OpenBSD 6.9 with this IrrlichtMt branch. I am not running a graphical environment, and I think will probably have to set one up before we know that this works properly, am I right in this thinking?

include/IrrCompileConfig.h
```diff
@@ -132,6 +132,10 @@
 #endif
 #endif
 
+#if defined(__FreeBSD__) || defined(__OpenBSD__)
+#define NO_IRR_COMPILE_WITH_JOYSTICK_EVENTS_
+#endif
```
sfan5, line 137: Can better be put as an ifdef around this line:
grafik

JosiahWI: I tried something like that, but I may have to undo it because I should've asked for clarification first. I'm confused about using an ifdef because it would basically need an ifndef, and I'm not sure that can or should replace defined(), based on the rest of the code which I tried to match the structure of as closely as possible.

sfan5: The way it is now is correct, not sure what the question is.

JosiahWI:
It looks like I may be unable to test Minetest graphically because setting up a desktop on my OpenBSD installation is not going well. OpenBSD isn't really intended as a graphical environment, which does raise the question of why this feature was requested.

PR #64, opened 2021-09-09
Readd TGA format support

rollerozxa:
This PR readds support for the TGA image format in IrrlichtMT, which was previously removed in PR #48 under the impression of it being unused.

Needs to be tested along with the [#11598](https://github.com/minetest/minetest/pull/11598) PR for the engine.

image

Mod containing a node with a TGA texture, for testing: [tga_test.zip](https://github.com/minetest/irrlicht/files/7135017/tga_test.zip)


include/IrrCompileConfig.h
```diff
@@ -411,6 +416,11 @@ B3D, MS3D or X meshes */
 #ifdef NO_IRR_COMPILE_WITH_PNG_WRITER_
 #undef _IRR_COMPILE_WITH_PNG_WRITER_
 #endif
+//! Define _IRR_COMPILE_WITH_TGA_WRITER_ if you want to write .tga files
+#define _IRR_COMPILE_WITH_TGA_WRITER_
+#ifdef NO_IRR_COMPILE_WITH_TGA_WRITER_
```
sfan5, line 421: The TGA writer is truly unused, don't add it back.

Wuzzy2:
It seems that the supported image file formats are NOT documented in `lua_api.txt` yet. I suggest to add this info, if this is not out of scope for this PR.

Anyway, it was a mistake to remove the long-standing TGA support for absolutely no reason. I have reports of people who depended on this. How did TGA removal manage to sneak through QA?

include/IrrCompileConfig.h
rollerozxa: Alright, removed it.

sfan5:
> How did TGA removal manage to sneak through QA?

It was considered obscure enough that nobody would or should be using it (no proper compression).
Even after @AFCMS mentioned that MCL needs them they didn't say anything else after being pointed to BMP or the (now implemented) PNG encoder API.
2 months later (today), erlehmann [dropped by in -dev](https://irc.minetest.net/minetest-dev/2021-09-09#i_5872574) to explain why it was in fact a bigger problem.

> I have reports of people who depended on this.

That's great. Why do we not have these reports?

AFCMS:
> Even after @AFCMS mentioned that MCL needs them they didn't say anything else after being pointed to BMP or the (now implemented) PNG encoder API.

The PNG encoder CAN be used in mcl2, but will break compatibility with older maps.
Also, mcl2 *aims* to support the two last versions of mt and this will of course cause a big problem.

appgurueu:
> That's great. Why do we not have these reports?

Because this was a rather obscure  and large PR not even in the main repo, but in the irrlicht repo, seemingly only doing code cleanup. No effort was made to reach out to mod devs or conduct actual usage data; fallacious assumptions were made.

erlehmann:
@appgurueu created a script to spider contentdb for additional file formats that might have been supported and are not supported right now. @appgurueu could you link the script and the results here maybe?

rollerozxa:
[Here is the Gist with the results](https://gist.github.com/appgurueu/63ffccbdd72c18506dc22d5206394de1) but they are irrelevant for this PR since it doesn't show any TGA files being present in Minetest mods or games source code.  The usage of TGA files is for programmatically creating images, which can't be shown with a file extension analysis.

erlehmann:
@rollerozxa I just wanted it to be here for the record. Thanks!

PR #77, opened 2021-11-08
Fix updating of vertex normals for animated meshes

x2048:
This PR makes CMeshManipulator::recalculateNormals() work on ISkinnedMesh implementations.

Two parts of the PR are:
* Add ISkinnedMesh::updateJointCache() / CSkinnedMesh::updateJointCache() method to copy vertex positions and normals into the weights of the animated joints
* Use updateJointCache() when CMeshManipulator detects an ISkinnedMesh implementation.

As a side effect, any manipulations on the mesh buffers when a skinned mesh is in static pose, followed by ISkinnedMesh::updateJointCache() will correctly update the animated mesh.

### DIFF
--- a/include/ISkinnedMesh.h
+++ b/include/ISkinnedMesh.h
@@ -80,7 +80,7 @@ namespace scene
 		virtual bool setHardwareSkinning(bool on) = 0;
 
 		//! Refreshes vertex data cached in joints such as positions and normals
-		virtual void refreshJointCache() {}
+		virtual void refreshJointCache() = 0;
 
 		//! A vertex weight
 		struct SWeight
--- a/include/ISkinnedMesh.h
+++ b/include/ISkinnedMesh.h
@@ -79,6 +79,9 @@ namespace scene
 		/* This feature is not implemented in Irrlicht yet */
 		virtual bool setHardwareSkinning(bool on) = 0;
 
+		//! Refreshes vertex data cached in joints such as positions and normals
+		virtual void refreshJointCache() {}
+
 		//! A vertex weight
 		struct SWeight
 		{
--- a/source/Irrlicht/CMeshManipulator.cpp
+++ b/source/Irrlicht/CMeshManipulator.cpp
@@ -151,8 +151,9 @@ void CMeshManipulator::recalculateNormals(scene::IMesh* mesh, bool smooth, bool 
 	for ( u32 b=0; b<bcount; ++b)
 		recalculateNormals(mesh->getMeshBuffer(b), smooth, angleWeighted);
 
-	if (mesh->getMeshType() == scene::E_ANIMATED_MESH_TYPE::EAMT_SKINNED) {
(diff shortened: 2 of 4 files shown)


include/ISkinnedMesh.h
```diff
@@ -79,6 +79,9 @@ namespace scene
 		/* This feature is not implemented in Irrlicht yet */
 		virtual bool setHardwareSkinning(bool on) = 0;
 
+		//! Refreshes vertex data cached in joints such as positions and normals
+		virtual void refreshJointCache() {}
```
sfan5, line 83: shouldn't this be `= 0;`?

source/Irrlicht/CMeshManipulator.cpp
```diff
@@ -149,6 +150,11 @@ void CMeshManipulator::recalculateNormals(scene::IMesh* mesh, bool smooth, bool
 	const u32 bcount = mesh->getMeshBufferCount();
 	for ( u32 b=0; b<bcount; ++b)
 		recalculateNormals(mesh->getMeshBuffer(b), smooth, angleWeighted);
+
+	if (mesh->getMeshType() == scene::E_ANIMATED_MESH_TYPE::EAMT_SKINNED) {
+		scene::ISkinnedMesh *smesh = reinterpret_cast<scene::ISkinnedMesh *>(mesh);
+		smesh->refreshJointCache();
+	}
```
sfan5, line 157:
```suggestion
	if (mesh->getMeshType() == EAMT_SKINNED)
	{
		ISkinnedMesh *smesh = (ISkinnedMesh *) mesh;
		smesh->refreshJointCache();
	}
```

source/Irrlicht/CSkinnedMesh.cpp
```diff
@@ -813,6 +813,21 @@ bool CSkinnedMesh::setHardwareSkinning(bool on)
...
+		SJoint *joint=AllJoints[i];
+		for (u32 j=0; j<joint->Weights.size(); ++j)
+		{
+			const u16 buffer_id=joint->Weights[j].buffer_id;
+			const u32 vertex_id=joint->Weights[j].vertex_id;
+			joint->Weights[j].StaticPos = LocalBuffers[buffer_id]->getVertex(vertex_id)->Pos;
+			joint->Weights[j].StaticNormal = LocalBuffers[buffer_id]->getVertex(vertex_id)->Normal;
+		}
+	}
+}
```
sfan5, line 830: Might need a call to `normalizeWeights`? Not sure.

x2048: normalizeWeights depends on joint strength, but not vertex normals or vertex positions. Looks like it's necessary as part of finalize(), but not here.

include/ISkinnedMesh.h
x2048: Indeed, I'm over complicating things :)

sfan5 (approved):

PR #79, opened 2021-11-10
Fix various GCC warnings

JosiahWI:
Minor code quality improvements. The main feature is that some unused functions were removed.


source/Irrlicht/CMakeLists.txt
```diff
@@ -264,6 +264,16 @@ add_library(IrrlichtMt)
...
 	IRRMESHOBJ IRROBJ IRRVIDEOOBJ
 	IRRIOOBJ IRROTHEROBJ IRRGUIOBJ)
+	target_compile_options(${object_lib}
+		PRIVATE
+			$<$<OR:$<CXX_COMPILER_ID:GNU>,$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
+				# The strlen optimization causes a stringop-overflow warning.
+				# The optimization caused no difference in the binary of
+				# a small reproducible case, suggesting that
+				# the warning is erroneous. However, I see no harm in
+				# disabling the optimization, just in case.
```
sfan5, line 274: I don't think disabling optimizations to get rid of mere warnings is harmless.

JosiahWI: You're right. I realized that solution was a lazy workaround and I've now fixed the warning properly, by improving the code quality.

JosiahWI:
There are a number of warnings in the Mac OS CI build yet. Some of these aren't as clean to fix nicely - I'm looking at the hidden virtual method warnings. The same warnings affect OpenBSD (which we users on) and presumably other BSD variants. I am happy to fix these, but I do not want to burden the devs or go against the project goals. Please advise. Thank you!

JosiahWI:
Clang issues those warnings also.

JosiahWI:
I am not planning to add any more new fixes. This is everything that I can test on my Debian dev workstation.

sfan5:
Can you split this into two commits? One that only removes unused code and one with the rest.

JosiahWI:
Is there a convenient way to do this in git or do I have to manually recommit the changes? I will force push once the branch tip is re-organized.

sfan5:
If you have commits separated so that each would fit into *either* of the final commits you can use `git rebase -i`, otherwise just recommit (I can recommend `git commit -p` for picking changes).

sfan5:
d4119ba6640fa96eb0587b46c4a64c2e818818a8 6d133e1bcc82c8646e0f94132a7e2cf09c840745

PR #83, opened 2021-12-11
use SSE registers for fp operations [affects mintest & irrlict] (fixes minetest issue 11810)

neoh4x0r:
This pull request fixes a minetest issue dealting with floating-point operations and it affects both minetest and irrlicht.
See: https://github.com/minetest/minetest/issues/11810

On a x86 system (could affect other platforms that handle floats this way) the x87 fp unit
stores floats with 80-bit precision, but when those floats are moved to other registers they suffer from
loss of precision and double rounding.

see
https://gcc.gnu.org/wiki/FloatingPointMath 
https://gcc.gnu.org/wiki/x87note

The solution to correct those two issues: (1) loss of precision and (2) double rounding, is to have all floating-point operations occur in sse registers.

That solution can be enabled by adding the following cmake options (in CMakeLists.txt):
```
add_compile_options(-mfpmath=sse -msse2)
```

This definately fixes the issue on i386 (the original issue involved Debian 10 Buster) but should affect all i386 and later systems.
On x86_64 (amd64) the issue wasn't present and the implemented solution (above) didn't affect the outcome.

The only potential pitfall (by enabling sse registers by default) would be on systems that don't have sse support.
However, I don't know if that is really an issue today (cpus that don't have sse support).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,11 @@ if(NOT CMAKE_BUILD_TYPE)
 	set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type: Debug or Release" FORCE)
 endif()
 
+# use SSE for floating-point operations to avoid issues with improper fp-rounding and loss of precision
+# when moving fp-data to incompatible or less-precise registers/storage locations
+# see https://gcc.gnu.org/wiki/FloatingPointMath and https://gcc.gnu.org/wiki/x87note
+add_compile_options(-mfpmath=sse -msse2)
+
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 add_subdirectory(source/Irrlicht)
 


JosiahWI:
https://stackoverflow.com/questions/6901369/how-to-detect-sse-availability-in-cmake

I just looked for this, and I think the accepted (and only at this time) answer to this SO post is quite good. Writing our own script to test for SSE instructions is probably the best way to do this, although if we don't want to put that much time in copying an existing script would be a lower quality but quicker way to have a test. If you don't want to write a test, I may be interested in doing it, but I do not have a lot of time, so I am not committed to this.

CMakeLists.txt
```diff
@@ -38,7 +38,9 @@ endif()
 # use SSE for floating-point operations to avoid issues with improper fp-rounding and loss of precision
 # when moving fp-data to incompatible or less-precise registers/storage locations
 # see https://gcc.gnu.org/wiki/FloatingPointMath and https://gcc.gnu.org/wiki/x87note
+IF(NOT (CMAKE_SYSTEM_PROCESSOR MATCHES "^arm" OR  CMAKE_SYSTEM_PROCESSOR MATCHES "^aarch64") )
 add_compile_options(-mfpmath=sse -msse2)
+ENDIF(NOT (CMAKE_SYSTEM_PROCESSOR MATCHES "^arm" OR  CMAKE_SYSTEM_PROCESSOR MATCHES "^aarch64") )
```
JosiahWI: Please use lowercase if\endif and indent the contained block, to be consistent with the style of the rest of the build system.

JosiahWI (changes requested):

CMakeLists.txt
neoh4x0r: @JosiahWI Yeah using lowercase to match coding styling would be a trivial thing to do.
However, @sfan5  has suggested an alternative which doesn't require writing a script (and uses lower-case to match style).

The question would be which is the correct CMakeLists.txt file to put it.
```
/CMakeLists.txt
/source/Irrlicht/CMakeLists.txt
```
-----

Diff from @sfan5
See https://github.com/minetest/minetest/pull/11853#issuecomment-991896648

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4803b475b..c3fec8822 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -759,6 +759,15 @@ else()
                endif()
        endif()
 
+       # Enable SSE for floating point math on 32-bit x86 by default
+       # reasoning see issue #11810 and https://gcc.gnu.org/wiki/FloatingPointMath
+       if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+               check_c_source_compiles("#ifndef __i686__\n#error\n#endif\nint main(){}" IS_I686)
+               if(IS_I686)
+                       set(OTHER_FLAGS "${OTHER_FLAGS} -mfpmath=sse -msse")
+               endif()
+       endif()
+
        if(MINGW)
                set(OTHER_FLAGS "${OTHER_FLAGS} -mthreads -fexceptions")
                set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWIN32_LEAN_AND_MEAN")
```

JosiahWI: If we put this in /CMakeLists.txt it will be clear that it applies to every target after it is defined, so I think that is a good place to put it. If it were put in /source/Irrlicht/CMakeLists.txt for example, I believe it may or may not affect my unit test work at all depending on which subdirectory gets added first.

neoh4x0r: How should the flags be set?
In that diff the following is used:
```
set(OTHER_FLAGS "${OTHER_FLAGS} -mfpmath=sse -msse")
```

I guess it would be good to use this?
```
add_compile_flags(-mfpmath=sse -msse)
```
Or is there a specfic convention for setting compiler flags in irrlicht (like with minetest)

sfan5: https://github.com/minetest/irrlicht/blob/master/source/Irrlicht/CMakeLists.txt#L20

neoh4x0r: So that should be in source/Irrlicht/CMakeLists.txt

I added it just after the if block at line 20:
```
if(CMAKE_CXX_COMPILER_ID MATCHES "^(GNU|Clang|AppleClang)$")
	set(CMAKE_CXX_FLAGS_RELEASE "-O3")
	set(CMAKE_CXX_FLAGS_DEBUG "-g")

	add_compile_options(-Wall -pipe -fno-exceptions -fno-rtti)
elseif(MSVC)
	string(APPEND CMAKE_CXX_STANDARD_LIBRARIES " msvcrt.lib") # ???? fuck off

	add_compile_options(/GR- /Zl)
endif()

# Enable SSE for floating point math on 32-bit x86 by default
# reasoning see minetest issue #11810 and https://gcc.gnu.org/wiki/FloatingPointMath
if(CMAKE_SIZEOF_VOID_P EQUAL 4)
	check_c_source_compiles("#ifndef __i686__\n#error\n#endif\nint main(){}" IS_I686)
	if(IS_I686)
		message(STATUS "Detected Intel x86: using SSE instead of x87 FPU")
		add_compile_flags(-mfpmath=sse -msse)
	endif()
endif()
```

On Debian Buster the version of cmake is 3.13 (but 3.18 is available from backports).

It complained that **check_c_source_compiles** was an unknown command.
Is there something that needs to be set in order for cmake to use **check_c_source_compiles**

JosiahWI: You may need to include the CheckCSourceCompiles module first.

https://cmake.org/cmake/help/v3.5/module/CheckCSourceCompiles.html

neoh4x0r: I neeed to add this as well:
```
enable_language(C)
include(CheckCSourceCompiles)
```

But now it complains that **add_compile_options** is an unknown command.
does that need an include or other command to enable?

neoh4x0r: I needed to add the following:

```
enable_languag(C)
include(CheckCSourceCompiles)
```

Also add_compile_flags should have been add_compile_options

neoh4x0r:
I know the minetest sets **/arch:SSE** for the MSVC compiler (for release-builds, but it's not clear about debug-builds).

Does Irrlicht need special handling to enable the SSE support on 32-bit Windows (like above) ?

source/Irrlicht/CMakeLists.txt
```diff
@@ -24,6 +24,18 @@ elseif(MSVC)
...
+# reasoning see minetest issue #11810 and https://gcc.gnu.org/wiki/FloatingPointMath
+if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+	enable_language(C)
+	include(CheckCSourceCompiles)
+	check_c_source_compiles("#ifndef __i686__\n#error\n#endif\nint main(){}" IS_I686)
+	if(IS_I686)
+		message(STATUS "Detected Intel x86: using SSE instead of x87 FPU")
+		add_compile_options(-mfpmath=sse -msse)
+	endif()
+endif()
```
sfan5, line 37: This needs to be inside the `if(CMAKE_CXX_COMPILER_ID ...`

You can add this inside the other branch just like Minetest:
<pre>
if(CMAKE_SIZEOF_VOID_P EQUAL 4)
    add_compile_options(/arch:SSE)
endif()
</pre>

neoh4x0r: I moved the sse fix into the CMAKE_CXX_COMPILER_ID if block.

Also, does **/arch:SSE** need to be enabled on MSVC for a non-intel cpu (such as. amd)

sfan5: Like I said both AMD and Intel are affected because they implement the same standard.

source/Irrlicht/CMakeLists.txt
```diff
@@ -18,10 +18,29 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "^(GNU|Clang|AppleClang)$")
 	set(CMAKE_CXX_FLAGS_DEBUG "-g")
 
 	add_compile_options(-Wall -pipe -fno-exceptions -fno-rtti)
+
+	# Enable SSE for floating point math on 32-bit x86 by default
+	# reasoning see minetest issue #11810 and https://gcc.gnu.org/wiki/FloatingPointMath
+	if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+		enable_language(C)
+		include(CheckCSourceCompiles)
+		check_c_source_compiles("#ifndef __i686__\n#error\n#endif\nint main(){}" IS_I686)
```
sfan5, line 27: Instead of the `enable_language(C)` you can also change this to use https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html

neoh4x0r: I also needed to change check_c_source_compiles to check_cxx_source_compiles

sfan5 (approved):

nerzhul (approved):

PR #85, opened 2021-12-19
Update CIrrDeviceSDL to support SDL2

emmauss:
This updates the sdl version CIrrDeviceSDL  uses from 1.2 to 2.0. Also, the compiler config header now allows building with only sdl device on windows, x11 and osx.


source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -28,7 +27,7 @@
 #endif
 
 #ifdef _MSC_VER
-#pragma comment(lib, "SDL.lib")
+//#pragma comment(lib, "SDL2.lib")
```
sfan5, line 30: You can just remove this entirely

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -212,51 +211,27 @@ CIrrDeviceSDL::CIrrDeviceSDL(const SIrrlichtCreationParameters& param)
...
 #if defined(_IRR_WINDOWS_)
-		SDL_putenv("SDL_VIDEODRIVER=directx");
+		SDL_setenv("SDL_VIDEODRIVER","directx", 1);
 #elif defined(_IRR_OSX_PLATFORM_)
-		SDL_putenv("SDL_VIDEODRIVER=Quartz");
+		SDL_setenv("SDL_VIDEODRIVER", "Quartz", 1);
 #elif !defined(_IRR_EMSCRIPTEN_PLATFORM_)
-		SDL_putenv("SDL_VIDEODRIVER=x11");
+		SDL_setenv("SDL_VIDEODRIVER", "x11", 1);
 #endif
```
sfan5, line 219: Hardcoding these seems unnecessary, can this be removed and it still works?

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -212,51 +211,27 @@ CIrrDeviceSDL::CIrrDeviceSDL(const SIrrlichtCreationParameters& param)
...
 
 	if ( CreationParams.Fullscreen )
-		SDL_Flags |= SDL_FULLSCREEN;
+		SDL_Flags |= SDL_WINDOW_FULLSCREEN;
 	else if ( Resizable )
-		SDL_Flags |= SDL_RESIZABLE;
+		SDL_Flags |= SDL_WINDOW_RESIZABLE;
 	if (CreationParams.DriverType == video::EDT_OPENGL)
-		SDL_Flags |= SDL_OPENGL;
-	else if (CreationParams.Doublebuffer)
```
sfan5, line 256: should the doublebuffer option result in a call `SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 0)` (or similar) or is that already the default?

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -712,25 +703,30 @@ bool CIrrDeviceSDL::run()
...
+				{
+					Width = SDL_event.window.data1;
+					Height = SDL_event.window.data2;
 #ifdef _IRR_EMSCRIPTEN_PLATFORM_
-				Screen = SDL_SetVideoMode( 0, 0, 32, SDL_Flags ); // 0,0 will use the canvas size
+					//SDL_CreateWindowAndRenderer(0, 0, SDL_Flags, &Window, &Renderer); // 0,0 will use the canvas size
 #else //_IRR_EMSCRIPTEN_PLATFORM_
- 				Screen = SDL_SetVideoMode( Width, Height, 0, SDL_Flags );
+ 					//Window = SDL_CreateWindow("", SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, Width, Height, SDL_Flags);
 #endif //_IRR_EMSCRIPTEN_PLATFOR
```
sfan5, line 724: lines should be removed if no longer needed
(also applies in a few more places further down)

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -1019,20 +1018,20 @@ video::IVideoModeList* CIrrDeviceSDL::getVideoModeList()
...
-				{
-					for (u32 i=0; modes[i]; ++i)
-						VideoModeList->addMode(core::dimension2d<u32>(modes[i]->w, modes[i]->h), vi->vfmt->BitsPerPixel);
+			int display_modes = SDL_GetNumDisplayModes(0);
+			if (display_modes != 0) {
+				for (u32 i = 0; i < display_modes; ++i) {
+					SDL_DisplayMode *mode;
+					if (SDL_GetDisplayMode(0, i, mode) > 0) {
+
+						VideoModeList->addMode(core::dimension2d<u32>(mode->w, mode->h),
```
sfan5, line 1027: You can delete all code relating to video modes.

source/Irrlicht/CIrrDeviceSDL.h
```diff
@@ -215,7 +215,9 @@ namespace irr
 			CIrrDeviceSDL* Device;
 			core::position2d<s32> CursorPos;
 			bool IsVisible;
-		};
+		};
+
+	SDL_Window* Window;
```
sfan5, line 220: This can be private even if it's used in the cursor control class.

source/Irrlicht/CMakeLists.txt
```diff
@@ -79,6 +80,14 @@ if(XINPUT2_ENABLED)
 	find_library(XINPUT_LIBRARY Xi REQUIRED)
 endif()
 
+check_symbol_exists(_IRR_COMPILE_WITH_SDL_DEVICE_ "IrrCompileConfig.h" SDL_ENABLED)
+
+unset(SDL_LIBRARIES CACHE)
```
sfan5, line 85: unset cache is incorrect here I think
maybe `set(SDL_LIBRARIES "")`?

source/Irrlicht/CMakeLists.txt
```diff
@@ -114,6 +124,8 @@ set(link_libs
 	"${ZLIB_LIBRARY}"
 	"${JPEG_LIBRARY}"
 	"${PNG_LIBRARY}"
+	"${SDL2_LIBRARY}"
```
sfan5, line 127: could add this one to `SDL_LIBRARIES` in the first place

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -288,8 +288,7 @@ bool COpenGLDriver::beginScene(u16 clearFlag, SColor clearColor, f32 clearDepth,
 		ContextManager->activateContext(videoData, true);
 
 #if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
-	if ( DeviceType == EIDT_SDL )
```
sfan5, line 291: The if's here have to stay

sfan5 (commented):
Thanks for contributing this, up-to-date SDL support is very useful

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -843,13 +839,14 @@ bool CIrrDeviceSDL::activateJoysticks(core::array<SJoystickInfo> & joystickInfo)
...
-		Joysticks.push_back(SDL_JoystickOpen(joystick));
+		SDL_Joystick *opened_joystick = SDL_JoystickOpen(joystick);
+		Joysticks.push_back(opened_joystick);
 		SJoystickInfo info;
 
 		info.Joystick = joystick;
 		info.Axes = SDL_JoystickNumAxes(Joysticks[joystick]);
 		info.Buttons = SDL_JoystickNumButtons(Joysticks[joystick]);
-		info.Name = SDL_JoystickName(joystick);
+		info.Name = SDL_JoystickName(opened_joystick);
```
ShadowNinja, line 849: This should be consistent with the other lines here, they should all use either `opened_joystick` or `Joysticks[joystick]`.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -1048,24 +1047,25 @@ void CIrrDeviceSDL::setResizable(bool resize)
...
+		if ( SDL_Flags & SDL_WINDOW_OPENGL )
 		{
 			// For unknown reasons the hack with sharing resources which was added in Irrlicht 1.8.5 for this no longer works in 1.9
 			// But at least we got a new WindowResizable flag since Irrlicht 1.9.
 			os::Printer::log("setResizable not supported with this device/driver combination. Use SIrrCreationParameters.WindowResizable instead.", ELL_WARNING);
 			return;
 		}
 #endif
-
+		*/
```
ShadowNinja, line 1060: Remove instead of just commenting out.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -1122,15 +1124,16 @@ bool CIrrDeviceSDL::isWindowActive() const
 			return false;
 	}
 #endif
-	const Uint8 appState = SDL_GetAppState();
-	return (appState&SDL_APPACTIVE && appState&SDL_APPINPUTFOCUS) ? true : false;
+	const u32 appState = SDL_GetWindowFlags(Window);
+	return (appState & SDL_WINDOW_SHOWN && appState&SDL_WINDOW_INPUT_FOCUS) ? true : false;
```
ShadowNinja, line 1128:
```suggestion
	const u32 windowFlags = SDL_GetWindowFlags(Window);
	return (windowFlags & SDL_WINDOW_SHOWN) && (windowFlags & SDL_WINDOW_INPUT_FOCUS);
```

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -1122,15 +1124,16 @@ bool CIrrDeviceSDL::isWindowActive() const
...
 }
 
 
 //! returns if window has focus.
 bool CIrrDeviceSDL::isWindowFocused() const
 {
-	return (SDL_GetAppState()&SDL_APPINPUTFOCUS) ? true : false;
+	const u32 appState = SDL_GetWindowFlags(Window);
+	return (appState & SDL_WINDOW_INPUT_FOCUS) ? true : false;
```
ShadowNinja, line 1136:
```suggestion
	const u32 windowFlags = SDL_GetWindowFlags(Window);
	return windowFlags & SDL_WINDOW_INPUT_FOCUS;
```

ShadowNinja (commented):
Very cool, SDL2 is probably our best path forward for native Wayland support.  I noted a few suggestions.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -557,10 +554,14 @@ bool CIrrDeviceSDL::run()
 
 			postEventFromUser(irrevent);
 			break;
-
+		case SDL_MOUSEWHEEL:
+			irrevent.EventType = irr::EET_MOUSE_INPUT_EVENT;
+			irrevent.MouseInput.Event = irr::EMIE_MOUSE_WHEEL;
+			irrevent.MouseInput.Wheel = SDL_event.wheel.y > 0 ? 1.0f : -1.0f;
```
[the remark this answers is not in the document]
v-rob: `preciseY` is only available since SDL 2.0.18, which is pretty recent, as an FYI.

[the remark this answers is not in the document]
ShadowNinja: Oh, that's too recent to depend on, but it could be in an `#if SDL_VERSION_ATLEAST(2, 0, 18)` block.

emmauss:
Address above reviews

source/Irrlicht/CIrrDeviceSDL.cpp
emmauss: SDL2 by default is double buffered. SDL_GL_DOUBLEBUFFER is boolean, so I guess I should set the attribute to false if single buffering is required

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -288,7 +288,7 @@ bool COpenGLDriver::beginScene(u16 clearFlag, SColor clearColor, f32 clearDepth,
 		ContextManager->activateContext(videoData, true);
 
 #if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
-	if ( DeviceType == EIDT_SDL )
+	if (SDLDevice->getType() == EIDT_SDL)
```
sfan5, line 291: This should be `if (SDLDevice)` to properly check for its presence

include/IrrCompileConfig.h
```diff
@@ -47,7 +47,7 @@
 //! Example: NO_IRR_COMPILE_WITH_X11_ would disable X11
 
 //! Uncomment this line to compile with the SDL device
-//#define _IRR_COMPILE_WITH_SDL_DEVICE_
+#define _IRR_COMPILE_WITH_SDL_DEVICE_
```
sfan5, line 50: If you want you can add a CI run that compiles and tests this, but it'll have to stay disabled by default.

emmauss: I forgot this was enabled. I meant to keep it commented out.

emmauss:
Addressed comments

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -309,11 +309,8 @@ bool COpenGLDriver::endScene()
 		status = ContextManager->swapBuffers();
 
 #ifdef _IRR_COMPILE_WITH_SDL_DEVICE_
-	if ( DeviceType == EIDT_SDL )
-	{
-		SDL_GL_SwapBuffers();
-		status = true;
-	}
+	SDLDevice->SwapWindow();
```
sfan5, line 312: `if (SDLDevice)` here too :)

emmauss: Done

ShadowNinja:
I had to apply the following patch to get it to compile:
```diff
diff --git a/source/Irrlicht/COpenGLCommon.h b/source/Irrlicht/COpenGLCommon.h
index d8ecc2e..eff48b6 100644
--- a/source/Irrlicht/COpenGLCommon.h
+++ b/source/Irrlicht/COpenGLCommon.h
@@ -41,8 +41,12 @@
                #define GL_GLEXT_PROTOTYPES 1
                #define GLX_GLXEXT_PROTOTYPES 1
        #endif
-       #include <SDL/SDL_video.h>
-       #include <SDL/SDL_opengl.h>
+       #include <SDL2/SDL_video.h>
+       #include <SDL2/SDL_opengl.h>
+       #include <GL/gl.h>
+       #if defined(_IRR_OPENGL_USE_EXTPOINTER_)
+       #include <GL/glext.h>
+       #endif^M
 #else
        #if defined(_IRR_OPENGL_USE_EXTPOINTER_)
                #define GL_GLEXT_LEGACY 1
```
After that I'm able to launch Minetest, however there are some issues:

* Keys aren't read properly. A lot of keys (e.g. shift and the arrow keys) seem to be read as tab.
* When running under native Wayland (`SDL_VIDEODRIVER=wayland`) relative mouse movement is bugged (moving the mouse slightly to the left and stopping will cause the view to continuously move left until you move the mouse back to the right)

emmauss:
> I had to apply the following patch to get it to compile:
> 
> ```diff
> diff --git a/source/Irrlicht/COpenGLCommon.h b/source/Irrlicht/COpenGLCommon.h
> index d8ecc2e..eff48b6 100644
> --- a/source/Irrlicht/COpenGLCommon.h
> +++ b/source/Irrlicht/COpenGLCommon.h
> @@ -41,8 +41,12 @@
>                 #define GL_GLEXT_PROTOTYPES 1
>                 #define GLX_GLXEXT_PROTOTYPES 1
>         #endif
> -       #include <SDL/SDL_video.h>
> -       #include <SDL/SDL_opengl.h>
> +       #include <SDL2/SDL_video.h>
> +       #include <SDL2/SDL_opengl.h>
> +       #include <GL/gl.h>
> +       #if defined(_IRR_OPENGL_USE_EXTPOINTER_)
> +       #include <GL/glext.h>
> +       #endif^M
>  #else
>         #if defined(_IRR_OPENGL_USE_EXTPOINTER_)
>                 #define GL_GLEXT_LEGACY 1
> ```
> 
> After that I'm able to launch Minetest, however there are some issues:
> 
>     * Keys aren't read properly. A lot of keys (e.g. shift and the arrow keys) seem to be read as tab.
> 
>     * When running under native Wayland (`SDL_VIDEODRIVER=wayland`) relative mouse movement is bugged (moving the mouse slightly to the left and stopping will cause the view to continuously move left until you move the mouse back to the right)

Thanks for testing. I do not readily have a linux system, so testing on it skipped me. The header and keys issues should be fixed. I'm unable to test the mouse issue on wayland since wayland is not supported on the system I got to test linux.

ShadowNinja:
It looks like the issue is that Minetest just hides the cursor then warps it back to the center of the window on every frame.  Wayland doesn't let you warp the cursor, you have to lock the cursor to the window use relative mouse motion instead (which `SDL_SetRelativeMouseMode` does).

I remember seeing similar issues before on other platforms (I think it sometimes happened on Windows?) so it's something that we should fix, but fixing it involves making changes to Minetest's mouse handling, so it wouldn't be a part of this PR.

sfan5 (approved):
Unicode input doesn't work, the clipboard is nonfunctional and I doubt this runs on Windows due to the [SDL_main](https://stackoverflow.com/questions/11976084/why-sdl-defines-main-macro) mechanism, but it's a start.
I'll merge this later.

v-rob:
Glad to see something going through.  I still do want to use SDL2 for everything if possible, so I'll likely work on this myself sometime.

PR #86, opened 2021-12-22
Remove unused attribute saving and loading

rollerozxa:
Irrlicht contains an attributes system which is like a typed key-value store. It is used for the following functionalities:
1. Debugging the scene manager. (disabled by default for performance reasons, requires `_IRR_SCENEMANAGER_DEBUG` to be set)
2. Storing driver properties. (unused? haven't touched it though)
3. Saving and loading GUIs, materials and scenes to and from XML files.

The third functionality has Minetest never used, and some code for it was even removed in the IrrlichtMt commit 729c214c003efb738c08f27817d569feb318dbf5 removing XML support and functionality relying on it. However, there were still a lot of methods left related to it, for (de)serializing GUI elements, scene nodes and materials into/from attributes. These methods has been sitting unused with no references to them from anywhere else in the code in both IrrlichtMt and Minetest ever since 729c214c003efb738c08f27817d569feb318dbf5, so they should be completely safe to remove.

**This also spans into the Minetest engine repository, [#11881](https://github.com/minetest/minetest/pull/11881) needs to be merged together with this IrrlichtMt PR!**


nerzhul (approved):

sfan5:
You could delete the `ISceneUserDataSerializer` too.

rollerozxa:
Yeah I'll go ahead and do that. SceneParameters in particular would need some more changes in the engine code once again, so I'll put that in #11881 too.

rollerozxa:
I trimmed down CAttributes and related to only what is currently used for scene manager debugging and driver properties, but not completely removing it. Should I remove that too? Driver properties seem to be mostly unused, scene manager debugging can actually be useful but it can't be viewed at the moment even when compiled in.

sfan5:
Keep those. Disregard my earlier suggestion, it's better if this PR restricts itself to removing gui/scene/material property stuff only.

rollerozxa:
Added back non-debug related SceneParameters, added back the boolean attribute type and reverted the changes in the Minetest PR.

sfan5 (approved):
Actual line changes were verified with `git diff --diff-algorithm=minimal HEAD FETCH_HEAD | grep '^\+'`
Tested with the respective engine PR.

PR #90, opened 2022-01-01
Reset mesh animation state before recalculating normals

### LINKED ISSUE #88: #77 breaks loading of some models (notably most NSSM models)
Turns out that #77 breaks model loading for some models (most NSSM models for example).
I'd suggest we revert that one until we figure out what the problem is.

@x2048 in case you wanted to have a look.

x2048:
Fixes #88 and minetest/minetest#11908.

Reset mesh to static position before calculating normals. This guarantees using correct vertex positions even if the mesh has been previously animated.

How to test
=========

* Build minetest/minetest#11747
* Install nssm mod with dependencies.
* Create a Minetest Game world and enable nssm mod and dependencies.
* Start the world and spawn a few Mantis mobs.
* Mobs should keep their shape and animation when new mobs are spawned.

### DIFF
--- a/include/ISkinnedMesh.h
+++ b/include/ISkinnedMesh.h
@@ -82,6 +82,9 @@ namespace scene
 		//! Refreshes vertex data cached in joints such as positions and normals
 		virtual void refreshJointCache() = 0;
 
+		//! Moves the mesh into static position.
+		virtual void resetAnimation() = 0;
+
 		//! A vertex weight
 		struct SWeight
 		{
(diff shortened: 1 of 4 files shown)


lhofhansl:
+1

I can confirm this works now. And the entities are shaded correctly too.

Thanks @x2048 for having a look!

lhofhansl:
So with this we do not need minetest/minetest#11914, right?

x2048:
Correct. I kept that one open just because it makes the code look more logical (load, correct normals, add to scene).

sfan5 (approved):

PR #91, opened 2022-01-01
Create OpenGL context using glXCreateContextAttribsARB

x2048:
Adds basic compatibility of Minetest with RenderDoc under Linux.

Allows capturing frames, but does not support replay.

This PR is ready for review.

How to test
=========

* Start Minetest and launch any world - Minetest should work as usual
* Start RenderDoc and launch Minetest under it - observe the frames being captured for examination

### DIFF
--- a/source/Irrlicht/CGLXManager.cpp
+++ b/source/Irrlicht/CGLXManager.cpp
@@ -320,7 +320,21 @@ bool CGLXManager::generateContext()
 		if (GlxWin)
 		{
 			// create glx context
-			context = glXCreateNewContext((Display*)CurrentContext.OpenGLLinux.X11Display, (GLXFBConfig)glxFBConfig, GLX_RGBA_TYPE, NULL, True);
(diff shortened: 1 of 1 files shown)


sfan5:
> Makes Minetest compatible with RenderDoc under Linux

Well not really. I also looked into this and made the same change locally.
IIRC RenderDoc manages to capture calls but none of the replaying capabilities work because of all the legacy calls. It also complained about the context not being core profile but enabling that would break all the legacy GL we still use.
Does this match your results?

x2048:
Sounds familiar. I could only use basic analysis tools in RenderDoc last time I used it. Without the change it would not even capture anything.

sfan5:
I'll revisit this after 5.5 and compare it to the implementation of the same thing I wrote.

sfan5:
pushed my implementation: d59bcdbd0706ac5eb90f55e346904315a9eaf3e2

PR #93, opened 2022-01-21
Add hash for vector2d

RichardTry:
Added support for vector2d in std::hash. This allows using vector2d as a key in std::unordered_map (especially for storing MapSectors in it).

### DIFF
--- a/include/vector2d.h
+++ b/include/vector2d.h
@@ -5,6 +5,7 @@
 #ifndef __IRR_POINT_2D_H_INCLUDED__
 #define __IRR_POINT_2D_H_INCLUDED__
 
+#include <functional>
 #include "irrMath.h"
 #include "dimension2d.h"
 
@@ -414,5 +415,20 @@ public:
 } // end namespace core
 } // end namespace irr
 
+namespace std
+{
+
+template<class T>
+struct hash<irr::core::vector2d<T> >
+{
+	size_t operator()(const irr::core::vector2d<T>& vec) const
+	{
+		size_t h1 = hash<T>()(vec.X);
+		size_t h2 = hash<T>()(vec.Y);
+		return (h1 << (4 * sizeof(h1)) | h1 >> (4 * sizeof(h1))) ^ h2;
+	}
+};
+
+}
 #endif
 


JosiahWI (commented):
This looks great! Has it been tested? I'm a little confused about the syntax for specializing a template for another template type.

EDIT: If you wrote test code for it, could it be added to the vector2d test in #84?

include/vector2d.h
```diff
@@ -5,6 +5,7 @@
 #ifndef __IRR_POINT_2D_H_INCLUDED__
 #define __IRR_POINT_2D_H_INCLUDED__
 
+#include <functional>
```
JosiahWI, line 8: Maybe include the STL header after the local headers? We don't pay a lot of attention to header order, but it would be a good practice.

JosiahWI (commented):
One more thing:

include/vector2d.h
```diff
@@ -414,5 +415,20 @@ class vector2d
...
+{
+
+template<class T>
+struct hash<irr::core::vector2d<T> >
+{
+	size_t operator()(const irr::core::vector2d<T>& vec) const
+	{
+		size_t h1 = hash<T>()(vec.X);
+		size_t h2 = hash<T>()(vec.Y);
+		return (h1 << (4 * sizeof(h1)) | h1 >> (4 * sizeof(h1))) ^ h2;
```
Desour, line 428: How good is this hash combination?
Have you thought about doing the same as boost:
https://github.com/boostorg/container_hash/blob/e9ffeceeab96ecd6979393676150874c3e5cd247/include/boost/container_hash/hash.hpp#L316

RichardTry:
@JosiahWI No, I didn't write a test for it. I just build IrrlichtMt with this and Minetest with unordered_map. Perhaps I could if you could explain me how.
@Desour I don't see any visible collisions, the code on the link does almost the same + some magic number.

include/vector2d.h
RichardTry: I don't see any visible collisions, the code on the link does almost the same + some magic number.

JosiahWI:
If you didn't write a test that's fine. I don't know whether anyone plans to review re-adding unit tests, and all the testing framework is on that PR still. If you wanted to add one it would involve modifying the Vector2D test (if I re-added it, I think I may have) in that PR to add another test function to those already there. If I didn't re-add it then it's a bit pointless haha.

sfan5:
Can you do this for `vector3d` too?

RichardTry:
> Can you do this for `vector3d` too?

Done.

PR #98, opened 2022-04-16
Add missing overrides

ndren:
Fixes warnings when compiling with `-Winconsistent-missing-override`.


x2048 (approved):

x2048:
Checked on CLang 10.0, looks good.

PR #99, opened 2022-04-26
Replace HWBufferMap with a list and back pointers

paradust7:
Fixes https://github.com/minetest/minetest/issues/12231#issue-1215956678

### DIFF
--- a/source/Irrlicht/CNullDriver.h
+++ b/source/Irrlicht/CNullDriver.h
@@ -21,7 +21,6 @@
 #include "SVertexIndex.h"
 #include "SLight.h"
 #include "SExposedVideoData.h"
-
 #include <list>
 
 namespace irr
--- a/source/Irrlicht/CNullDriver.h
+++ b/source/Irrlicht/CNullDriver.h
@@ -22,6 +22,8 @@
 #include "SLight.h"
 #include "SExposedVideoData.h"
 
+#include <list>
+
 namespace irr
 {
 namespace io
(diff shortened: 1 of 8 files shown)


source/Irrlicht/CNullDriver.h
```diff
@@ -388,25 +389,29 @@ namespace video
...
+				}
 			}
 
 			const scene::IMeshBuffer *MeshBuffer;
 			u32 ChangedID_Vertex;
 			u32 ChangedID_Index;
-			u32 LastUsed;
 			scene::E_HARDWARE_MAPPING Mapped_Vertex;
 			scene::E_HARDWARE_MAPPING Mapped_Index;
+			std::list<SHWBufferLink*>::iterator listPosition;
```
sfan5, line 414: did you take note of iterator invalidation?

sfan5 (approved):
Tested, looks like it works fine.

lhofhansl:
I find it hard to reason about whether HWBuffers are indeed released as soon as a mesh is not longer visible.
That was the crux of the original problem, HWBuffers would hang around for all cached meshes.

When the mesh's reference count is 0 that means the mesh is not visible?

sfan5:
It doesn't depend on the visibility of meshes but uses the lifetime of the mesh buffer.
When a hardware buffer is added it keeps an extra reference to the mesh, then when the reference count of the mesh drops to 1 both the hardware buffer and the last reference to the mesh are dropped.

lhofhansl:
Thanks @sfan5 . So that means that each _cached_ mesh will keep its HWBuffer around, instead of just the visible meshes?

In any case, I'll try it out when I get some time next.

sfan5:
> So that means that each _cached_ mesh will keep its HWBuffer around, instead of just the visible meshes?

Yes. This doesn't sound like a problem to me since the unload timeout on the client will get rid of old data at some point.

PR #101, opened 2022-05-02
Make irrArray backed by std::vector

paradust7:
2/2 of the fix to issue [#12251](https://github.com/minetest/minetest/issues/12251#issue-1221758115)


include/irrArray.h
```diff
@@ -131,106 +99,19 @@ class array
...
-		{
-			for (u32 i=0; i<used; ++i)
-				allocator.destruct(&data[i]);
-
-			allocator.deallocate(data); // delete [] data;
-		}
-		data = 0;
-		used = 0;
-		allocated = 0;
+		std::vector<T>().swap(m_data);
```
sfan5, line 111: `m_data.clear()`

include/irrArray.h
```diff
@@ -255,85 +135,51 @@ class array
...
-	void set_free_when_destroyed(bool f)
-	{
-		free_when_destroyed = f;
-	}
-
-
 	//! Sets the size of the array and allocates new elements if necessary.
 	/** Please note: This is only secure when using it with simple types,
 	because no default constructor will be called for the added elements.
 	\param usedNow Amount of elements now used. */
```
sfan5, line 146: comment can be removed

include/irrArray.h
```diff
@@ -549,11 +363,11 @@ class array
...
-		for (s32 i=used-1; i>=0; --i)
-			if (data[i] == element)
-				return i;
-
-		return -1;
+		auto it = std::find(m_data.rbegin(), m_data.rend(), element);
+		if (it == m_data.rend())
+			return -1;
+		size_t offset = it - m_data.rbegin();
+		return m_data.size() - offset - 1;
```
sfan5, line 370: Feel free to add basic unit tests [here](https://github.com/minetest/irrlicht/blob/master/examples/AutomatedTest/main.cpp), I can't tell at a glance whether this is correct.

include/irrArray.h
```diff
@@ -619,38 +409,29 @@ class array
...
+		m_data.swap(other.m_data);
+		std::swap(is_sorted, other.is_sorted);
+	}
+
+	//! Pull the contents of this array as a vector.
+	// The array is left empty.
+	std::vector<T> steal() {
+		std::vector<T> ret = std::move(m_data);
+		m_data.clear();
+		is_sorted = false;
```
sfan5, line 423: `is_sorted = true`

sfan5 (commented):
Not tested yet but looks good

include/irrArray.h
paradust7: I was conflicted about this one, because the comment says "Clears the array and deletes all allocated memory.". Generally clear() does not delete the allocated memory, it only resets the size to 0.
clear() followed by shrink_to_fit() would be more like it (although the spec allows the library to ignore shrink_to_fit).
This in-place swap is the usual way to guarantee a total free.

I will look at how array::clear() is used and see if there are any situations that rely on this behavior.

sfan5: In that case just add a comment so it's obvious why the code is that way :)

include/irrArray.h
paradust7: Added unit tests in another PR:

https://github.com/minetest/irrlicht/pull/103

Both the existing code and the new version pass.

include/irrArray.h
paradust7: done

include/irrArray.h
paradust7: added comment

include/irrArray.h
paradust7: done

sfan5 (approved):

sfan5:
You can delete `heapsort.h` too, it is only used in the old core::array code

PR #102, opened 2022-05-04
Enable GL_ALPHA_TEST in OpenGL driver for custom transparent alpha shaders

paradust7:
Fixes issue #11852


erlehmann:
Does it make sense to upstream this to irrlicht?

PR #103, opened 2022-05-06
Unit tests for irrArray

paradust7:
Basic unit tests for irr::core::array, added to irrlicht repo's AutomatedTest
Part of https://github.com/minetest/minetest/issues/12251


paradust7:
Not sure what to do about the \ alignment in test_helper. It was aligned in my editor, but there's no way to make it aligned on all displays, since tab widths vary.

paradust7:
MacOS failing because the compiler is not C++11.

sfan5 (approved):

PR #105, opened 2022-05-10
Remove core::list and replace uses with std::list

paradust7:
Biggest changes were made to IGUIElement and CGUITreeViewNode. I noticed they were doing linear searches for almost every operation touching the list. So I added a field (ParentPos) for both of them, to make list operations constant-time.

Part of https://github.com/minetest/minetest/issues/12312


source/Irrlicht/CGUITreeView.h
```diff
@@ -231,7 +230,8 @@ namespace gui
 		void*				Data;
 		IReferenceCounted*		Data2;
 		bool				Expanded;
-		core::list<CGUITreeViewNode*>	Children;
+		std::list<CGUITreeViewNode*>	Children;
+		std::list<CGUITreeViewNode*>::iterator ParentPos;
```
sfan5, line 234: can you add short comment(s) here too?

include/IGUIElement.h
```diff
@@ -308,17 +306,19 @@ class IGUIElement : virtual public IReferenceCounted, public IEventReceiver
...
-		for (; it != Children.end(); ++it)
-			if ((*it) == child)
-			{
-				(*it)->Parent = 0;
-				(*it)->drop();
-				Children.erase(it);
-				return;
-			}
+		assert(child->Parent == this);
+		Children.erase(child->ParentPos);
```
appgurueu, line 310: According to the [C++ ref](https://en.cppreference.com/w/cpp/container/list/erase) this might still be linear time due to the `erase` call. I also wonder how you keep the `ParentPos` of the potentially shifted children after this updated; wouldn't updating them require linear time as well?

sfan5 (approved):

PR #110, opened 2022-05-29
Fix line drawing: Explicitly draw both first & last pixel

appgurueu:
Fixes https://github.com/minetest/minetest/issues/11049

### DIFF
--- a/source/Irrlicht/COpenGLDriver.cpp
+++ b/source/Irrlicht/COpenGLDriver.cpp
@@ -1862,7 +1862,9 @@ void COpenGLDriver::draw2DLine(const core::position2d<s32>& start,
 
 		glDrawElements(GL_LINES, 2, GL_UNSIGNED_SHORT, Quad2DIndices);
 
-		// Draw non-drawn last pixel (search for "diamond exit rule")
+		// Draw sometimes non-drawn first & last pixel (search for "diamond exit rule")
+		// HACK this messes with alpha blending
+		glDrawArrays(GL_POINTS, 0, 0);
 		glDrawArrays(GL_POINTS, 1, 1);
 	}
 }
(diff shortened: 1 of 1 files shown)


source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -1862,7 +1862,9 @@ void COpenGLDriver::draw2DLine(const core::position2d<s32>& start,
 
 		glDrawElements(GL_LINES, 2, GL_UNSIGNED_SHORT, Quad2DIndices);
 
-		// Draw non-drawn last pixel (search for "diamond exit rule")
+		// Draw sometimes non-drawn first & last pixel (search for "diamond exit rule")
+		// HACK this messes with alpha blending
```
appgurueu, line 1866: Does it though? Testing with alpha values for the crosshair didn't yield anything; perhaps the blend mode is miraculously set correctly?

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -1862,7 +1862,9 @@ void COpenGLDriver::draw2DLine(const core::position2d<s32>& start,
 
 		glDrawElements(GL_LINES, 2, GL_UNSIGNED_SHORT, Quad2DIndices);
 
-		// Draw non-drawn last pixel (search for "diamond exit rule")
+		// Draw sometimes non-drawn first & last pixel (search for "diamond exit rule")
+		// HACK this messes with alpha blending
+		glDrawArrays(GL_POINTS, 0, 0);
```
luk3yx, line 1867: If I change the third parameter to `1`, the issue is fixed.

```suggestion
		glDrawArrays(GL_POINTS, 0, 1);

```

appgurueu: Should've read the gl docs instead of making the stupid assumption that these were start & end indices (both inclusive):

> void glDrawArrays( | GLenum | mode, | GLint | first, | GLsizei | count);

https://docs.gl/gl3/glDrawArrays

luk3yx (approved):
Thanks!

PR #112, opened 2022-06-08
Reset blend operation in 2d mode

x2048:
This fixes flicker in HUD when blend function is not GL_FUNC_ADD after 3D operations, for example, when rendering particles with blend = "sub" in minetest/minetest#11545 with changes from velartrill/minetest#3


sfan5:
Ah I forgot to ask, what about the OGLES1 and OGLES2 driver?

PR #114, opened 2022-06-21
Support Unicode characters properly in SDL2

ndren:
I was trying to look through on how SDL handles input and although some things work such as special characters in ASCII ($%) I have not been able to do the same for general wchar_t unicode characters.

This currently breaks all other in-game input (but also backspace) so is not usable though I would like to know if there is anything useful could come out of this approach.


ndren:
Thank you! I think I have found the issue that makes most characters broken; since I am casting the first item to wchar_t instead of converting the whole string, this only works for ASCII. However, [I could not find any portable way to convert string to wstring](https://stackoverflow.com/questions/2573834/c-convert-string-or-char-to-wstring-or-wchar-t).

ndren:
I figured it out! It wasn't that complicated, I just needed to remove duplicate functionality between the different input modes.

I have tested this and there are no more regressions.

ndren:
I should have said there are no more regressions that I knew of at the time... Backspace seems to be handled differently in the start menu compared to inside the game itself, so I've added a workaround for that.

Does the code look good overall?

sfan5:
This looks better functionality wise, but I have a few points:
* the reason why backspace needs a workaround should be investigated (something wrong in `KeyMap`?)
* `isNoUnicodeKey` looks unused now
* you should move the `case SDL_TEXTINPUT:` to be separate for better organization
* what about calling `SDL_StartTextInput` / `SDL_StopTextInput`. is this necessary / should this be left for later?

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -670,25 +615,16 @@ bool CIrrDeviceSDL::run()
...
-
-				if (convertCharToUppercase)
-					irrevent.KeyInput.Char = irr::core::locale_upper(irrevent.KeyInput.Char);
-
+				// These keys are handled differently in CGUIEditBox.cpp (may become out of date!)
+				// Control key is used in special character combinations, so keep that too
+				// Pass through the keysym only then so no extra text gets input
+				irrevent.KeyInput.Char = 0;
+				if (mp.SDLKey == KEY_DELETE || mp.SDLKey == KEY_RETURN || mp.SDLKey == KEY_BACK || irrevent.KeyInput.Control)
+					irrevent.KeyInput.Char = mp.SDLKey;
```
sfan5, line 627: do sdl keys roughly correspond to chars or why does this work?

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -648,6 +580,19 @@ bool CIrrDeviceSDL::run()
 			}
 			break;
 
+		case SDL_TEXTINPUT:
+			{
+				irrevent.EventType = irr::EET_STRING_INPUT_EVENT;
+				size_t size = strlen(SDL_event.text.text);
+				wchar_t tmp[size*2];
+				mbstowcs(tmp, SDL_event.text.text, size*2);
```
sfan5, line 588: there's a wrapper function for this btw
https://github.com/minetest/irrlicht/blob/master/include/irrString.h#L1431

source/Irrlicht/CIrrDeviceSDL.cpp
ndren: It does not work. (I thought it worked because I was testing the wrong menu. Now it works.)

ndren:
Should be good now! Thanks a lot for your help.

sfan5 (approved):
Works.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -670,25 +613,16 @@ bool CIrrDeviceSDL::run()
...
-				bool convertCharToUppercase =
-					irrevent.KeyInput.Shift != !!(SDL_event.key.keysym.mod & KMOD_CAPS);
-
-				if (convertCharToUppercase)
-					irrevent.KeyInput.Char = irr::core::locale_upper(irrevent.KeyInput.Char);
-
+				// These keys are handled differently in CGUIEditBox.cpp (may become out of date!)
+				// Control key is used in special character combinations, so keep that too
+				// Pass through the keysym only then so no extra text gets input
+				irrevent.KeyInput.Char = 0;
```
paradust7, line 623: This completely wipes out the case-sensitive ascii character (Key is case-insensitve, e.g. KEY_KEY_T) which means it's no longer possible to type text in the chat/command window. What's the intent here?

paradust7: Note: The existing code is (afaict) also very broken and many ascii characters fail to work in chat (because the KeyMap is wrong). I have a patch that fixes it for the minetest web version, but only supports ascii.

paradust7: Apologies, I just realized SDL on emscripten may be broken with regard to the SDL_TEXTINPUT event, which is the real issue here. I'll investigate further.

PR #115, opened 2022-06-27
Fix null dereferences reported by coverity

JosiahWI:
One of these was caused by a null check missing a return; the other was
a false positive, but I thought the code readability would benefit from
a little rewording, which should also quiet Coverity. There's a third one reported, but it's also clearly a false positive. I think I could simplify the logic and quiet that one as well, but I don't want to break the logic by mistake, so I'm leaving it out.

### DIFF
--- a/source/Irrlicht/CFileSystem.cpp
+++ b/source/Irrlicht/CFileSystem.cpp
@@ -783,7 +783,7 @@ EFileSystemType CFileSystem::setFileListSystem(EFileSystemType listType)
 //! Creates a list of files and directories in the current working directory
 IFileList* CFileSystem::createFileList()
 {
-	CFileList* r = 0;
+	CFileList* r = nullptr;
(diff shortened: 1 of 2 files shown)


source/Irrlicht/CFileSystem.cpp
```diff
@@ -861,6 +859,9 @@ IFileList* CFileSystem::createFileList()
 			}
 			closedir(dirHandle);
 		}
+		#else
+			// no native filesystem enabled
+			return nullptr;
```
sfan5, line 864: I don't think burying the return deep inside the function helps readability here

JosiahWI: That is a point. Maybe that one is better left how it was; do you want me to undo it?

sfan5: I'd prefer that yes

PR #117, opened 2022-07-04
Fix uninitialized color array in GUIListBox items

JosiahWI:
I timed the addition of a million list box items with and without the allocation, using the Timer class provided by learncpp.com:

```cpp
    gui::IGUIListBox* box{env->addListBox(rect<s32>{50, 140, 250, 210})};

    Timer t;
    for (int i{0}; i < 1000000; ++i) {
        box->addItem(L"yay", 0);
    }
    std::cout << "Time elapsed: " << t.elapsed() << " seconds\n";
```
    
I timed it 5 times with each version of the Irrlicht code, and took the average.
    
- without the zero initialization: 1.50901 seconds
- with the zero initialization: 1.54063 seconds

I was a little surprised by this result! But for the number of list box items that would make sense for an application, it should be negligible.


SmallJoker (approved):
LGTM

PR #130, opened 2022-08-17
Fixed type size, glHint and added more check null

GermanAizek:
@sfan5, review pr changes and feedback me.

### DIFF
--- a/source/Irrlicht/CB3DMeshFileLoader.cpp
+++ b/source/Irrlicht/CB3DMeshFileLoader.cpp
@@ -26,7 +26,7 @@ namespace scene
 
 //! Constructor
 CB3DMeshFileLoader::CB3DMeshFileLoader(scene::ISceneManager* smgr)
-: AnimatedMesh(0), B3DFile(0), NormalsInFile(false),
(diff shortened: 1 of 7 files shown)


GermanAizek:
@sfan5,
don't forget to do review this PR, thanks.

sfan5:
Which of these changes is a security improvement?

source/Irrlicht/CIrrDeviceWin32.cpp
```diff
@@ -1390,7 +1390,7 @@ void CIrrDeviceWin32::handleSystemMessages()
 
 	while (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
 	{
-		if (ExternalWindow && msg.hwnd == HWnd)
+		if (ExternalWindow)
```
sfan5, line 1393: this is a change in behaviour

GermanAizek:
> Which of these changes is a security improvement?

protection against null reference exception

source/Irrlicht/CIrrDeviceWin32.cpp
GermanAizek: fix it

sfan5 (approved):

PR #132, opened 2022-08-23
[nosquash] Implement X11 primary selection, and add clipboard support for SDL

Desour:
As #55 was merged, supporting primary selection is just a matter of using a different X11 Atom and using a separate string to store primary selection texts.
And this is exactly what the first commit does.

The second commit does exactly the same as the commit in the minetest repo PR, but with other member variable and utf8 to stringw function names.

## To Do

PR is ready for review.
Commits are best reviewed separately.

Don't squash.

## How to test

* Use X11.
* Place a sign in minetest.
* Mark text.
* Middle click.
* Do the same in other programs.

### DIFF
--- a/include/IOSOperator.h
+++ b/include/IOSOperator.h
@@ -26,11 +26,22 @@ public:
 	}
 
 	//! Copies text to the clipboard
+	//! \param text: text in utf-8
 	virtual void copyToClipboard(const c8* text) const = 0;
 
+	//! Copies text to the primary selection
+	//! This is a no-op on some platforms.
+	//! \param text: text in utf-8
+	virtual void copyToPrimarySelection(const c8* text) const = 0;
+
 	//! Get text from the clipboard
-	/** \return Returns 0 if no string is in there. */
+	//! \return Returns 0 if no string is in there, otherwise an utf-8 string.
 	virtual const c8* getTextFromClipboard() const = 0;
+
+	//! Get text from the primary selection
+	//! This is a no-op on some platforms.
+	//! \return Returns 0 if no string is in there, otherwise an utf-8 string.
+	virtual const c8* getTextFromPrimarySelection() const = 0;
 
 	//! Get the total and available system RAM
 	/** \param totalBytes: will contain the total system memory in Kilobytes (1024 B)
--- a/source/Irrlicht/COSOperator.cpp
+++ b/source/Irrlicht/COSOperator.cpp
@@ -19,6 +19,7 @@
 
 #if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
 #include <SDL_clipboard.h>
+#include <SDL_version.h>
 #elif defined(_IRR_COMPILE_WITH_X11_DEVICE_)
 #include "CIrrDeviceLinux.h"
 #endif
@@ -27,6 +28,19 @@
 #endif
 
 #include "fast_atof.h"
+
+#if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
+static bool sdl_supports_primary_selection = [] {
+#if SDL_VERSION_ATLEAST(2, 25, 0)
+	SDL_version linked_version;
+	SDL_GetVersion(&linked_version);
+	return (linked_version.major == 2 && linked_version.minor >= 25)
+			|| linked_version.major > 2;
+#else
+	return false;
+#endif
+}();
+#endif
 
 namespace irr
 {
@@ -111,7 +125,13 @@ void COSOperator::copyToPrimarySelection(const c8 *text) const
 	if (strlen(text)==0)
 		return;
 
-#if defined(_IRR_COMPILE_WITH_X11_DEVICE_)
+#if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
+#if SDL_VERSION_ATLEAST(2, 25, 0)
+	if (sdl_supports_primary_selection)
+		SDL_SetPrimarySelectionText(text);
+#endif
+
+#elif defined(_IRR_COMPILE_WITH_X11_DEVICE_)
     if ( IrrDeviceLinux )
         IrrDeviceLinux->copyToPrimarySelection(text);
 #endif
@@ -172,7 +192,19 @@ const c8* COSOperator::getTextFromClipboard() const
 //! gets text from the primary selection
 const c8* COSOperator::getTextFromPrimarySelection() const
 {
-#if defined(_IRR_COMPILE_WITH_X11_DEVICE_)
+#if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
+#if SDL_VERSION_ATLEAST(2, 25, 0)
+	if (sdl_supports_primary_selection) {
+		static char *text = nullptr;
+		if (text)
+			SDL_free(text);
+		text = SDL_GetPrimarySelectionText();
+		return text;
+	}
+#endif
+	return 0;
+
+#elif defined(_IRR_COMPILE_WITH_X11_DEVICE_)
     if ( IrrDeviceLinux )
         return IrrDeviceLinux->getTextFromPrimarySelection();
     return 0;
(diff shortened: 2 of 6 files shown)


Desour:
SDL now supports primary selection (on X11 and wayland; and inner-application on other platforms).

I've added calls to the sdl primary selection and clipboard functions.

I've tested the primary selection feature with both, loaded sdl version bigger and smaller than compiled version. The version check works correctly.

To use sdl, compile with `-DSDL_ENABLED=1` and uncomment this line:
https://github.com/minetest/irrlicht/blob/e9f205f952efac448974239526c5284134ae41b9/include/IrrCompileConfig.h#L48

Desour:
Rebased.

include/IOSOperator.h
```diff
@@ -26,12 +26,23 @@ class IOSOperator : public virtual IReferenceCounted
 	}
 
 	//! Copies text to the clipboard
+	//! \param text: text in utf-8
 	virtual void copyToClipboard(const c8* text) const = 0;
 
+	//! Copies text to the primary selection
+	//! This is a no-op on some platforms.
+	//! \param text: text in utf-8
+	virtual void copyToPrimarySelection(const c8* text) const = 0;
```
numberZero, line 35: Breaks binary compatibility, is that OK?

numberZero (commented):
Tested on X11, works fine. Code looks mostly OK.

include/IOSOperator.h
numberZero: Specifically, Minetest compiled against current HEAD would crash if ran against this version, because of VMT changes. If that is problematic, move the new virtual methods to the bottom like you would do for new fields.

Desour: TBH, I know almost nothing about binary compatibility in C++. Is it *guaranteed* that code that expect old vtables works with newer vtables if the new virtual member function is added at the bottom? If you know some references where such things are written down, I'd be very happy to read from them.

Also, I don't know if a binary compatibility break is ok. @sfan5 ?
And I haven't looked into how minetest uses its irrlicht revision versions.

sfan5: We don't care about binary compatibility. (each release is a higher [revision](https://github.com/minetest/irrlicht/commit/a0af653c3d1d0e4a17c9c9408ff7eb0fc697ad4b) and bumps the SOVERSION)
We'll stop caring about source compatibility soon too when the submodule stuff gets in.

ndren:
I have tested and the clipboard (ctrl-V) works fine on Sway (Wayland). I was a bit confused when I saw that the primary did not work, but I see now it's only implemented for X11.

This is not a blocker, but maybe the code could use the primary selection you added to SDL for posterity? (So that X11 -> SDL on X11 would not remove this feature.)

ndren:
Again, this is not required, and I'm happy for this to be left as-is. One clipboard is enough for me :)

Desour:
Primary selection should work with SDL, including on wayland.
Are you sure your SDL version is recent enough (i.e. >=2.26)?
Also, wayland compositors are not required to support primary selection (it's an extension). If yours doesn't, SDL doesn't support it either.

ndren:
I figured out my issue! Chat messages do not allow newlines, leading to my confusion when my clipboard did include newlines. Is there a nice way to gracefully handle this?

I'm thinking maybe either stripping newlines or maybe not pasting at all if newlines are present (I think I prefer keeping as much of the message as possible though.)

The complexity comes from other input fields allowing newlines, so I don't know what the best way to deal with this is. I'm not sure if this should be handled here, but I don't see other cases where newlines would unexpectedly appear.

ndren:
Both clipboards work when there are no newlines in the text, so it works fine otherwise.

sfan5:
[guiChatConsole](https://github.com/minetest/minetest/blob/master/src/gui/guiChatConsole.cpp) should handle this, you don't need to worry about it here.

Desour:
Oh, I've never actually tried pasting text with newline in the chat-console. Single-line editboxes allow this, afaik.
As the behaviour in this regard is the same as if you paste from clipboard selection, fixes for that can go into a separate PR.

source/Irrlicht/COSOperator.cpp
```diff
@@ -26,6 +29,19 @@
 
 #include "fast_atof.h"
 
+#if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
+static bool sdl_supports_primary_selection = [] {
+#if SDL_VERSION_ATLEAST(2, 25, 0)
+	SDL_version linked_version;
+	SDL_GetVersion(&linked_version);
+	return (linked_version.major == 2 && linked_version.minor >= 25)
+			|| linked_version.major > 2;
```
sfan5, line 38: this won't work.
by referencing the `SDL_SetPrimarySelectionText` symbol you already introduce a hard-dependency

source/Irrlicht/COSOperator.cpp
```diff
@@ -102,11 +119,36 @@ void COSOperator::copyToClipboard(const c8 *text) const
...
+}
+
+
 //! gets text from the clipboard
-//! \return Returns 0 if no string is in there, otherwise an utf-8 string.
 const c8* COSOperator::getTextFromClipboard() const
 {
-#if defined(_IRR_WINDOWS_API_)
+#if defined(_IRR_COMPILE_WITH_SDL_DEVICE_)
+	static char *text = nullptr;
```
sfan5, line 145: static variables are bad, put it into the class

sfan5 (commented):
actually should have reviewed this first...

source/Irrlicht/COSOperator.cpp
Desour: I've put `SDL_SetPrimarySelectionText` into an `#if` (see below), so it is only referenced if the compiled SDL version is new enough.

source/Irrlicht/COSOperator.cpp
Desour: See https://github.com/minetest/irrlicht/pull/180.

source/Irrlicht/COSOperator.cpp
sfan5: Yes, but the check I'm commenting on _doesn't do anything_.
If you compile with an old SDL, the code is never included.
If you compile with a new SDL, the code is included and but won't load if the runtime SDL is older so `sdl_supports_primary_selection` is always true.

Desour: I don't think I'm getting what you're trying to say.

If you compile with new SDL and use old runtime SDL, `sdl_supports_primary_selection` is false. `SDL_SetPrimarySelectionText` is in this case never loaded (as the code that does the symbol lookup is never reached).

Only if someone enables full relro (or if their OS does things differently), they'll only be able to launch minetest with a new runtime. But this not the default.

sfan5: Huh, I didn't even know this was a thing and thought that all symbols were resolved at startup. I certainly can't remember ever seeing a "unresolved symbol" error at _runtime_ in the 10(?) years I've been using Linux.
Regarding relro: Arch enables full relro, Ubuntu enables relro too (not sure if full), [Fedora has full relro too](https://www.redhat.com/en/blog/hardening-elf-binaries-using-relocation-read-only-relro) so in practice this code is still pretty pointless.

Desour: Well, yes, this code could probably be left away. But it doesn't hurt.
(And I'm certainly not going to implement a manual dlsym lookup for the destros shipping with full relro. x))
(Btw., FYI it's "symbol lookup error" and "undefined symbol" if it happens at runtime. (Just tried out.))

numberZero: > thought that all symbols were resolved at startup

On musl, that’s always the case. But on glibc, the default is to delay resolve until the first call (or at least that was the default last time I checked).

PR #133, opened 2022-09-15
Fix typo s/tranfrom/transform/

coldtobi:
This patch was provided form J. Puydt to Debian.
Origin: https://salsa.debian.org/games-team/minetest/-/blob/master/debian/patches/fix-typos.patch


sfan5:
Is this the only typo found in Irrlicht?

coldtobi:
This seems the only typo, at least the only one found by Debian tooling.
(There is another typo in minetest, but thats a different repo)

PR #134, opened 2022-09-22
Pass CFLAGS and CPPFLAGS to build process

coldtobi:
When overriding compiler options in cmake CMake does not by default honor environment set compiler options, this has to be manually considered. In Debian, CFLAGS and CPPFLAGS are injected by the environment to set some archive-wide defaults, for example enabling hardening features.

(If the environment variables are not set, this patch has no effect.)

Patch origin:
https://salsa.debian.org/games-team/minetest/-/blob/master/debian/patches/enable_hardening.patch (minus some extra blank lines, which I missed when creating the patch for Debian))


numberZero:
Does *any* CMake-based project does that?

numberZero:
Actually, this PR made me realize there is another problem. See #164 for details.

coldtobi:
> Does _any_ CMake-based project does that?

This is a known issue with CMake. Background: https://gitlab.kitware.com/cmake/cmake/-/issues/20631 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=653916

numberZero:
Can’t you simply `cmake -DCMAKE_C_FLAGS="$CFLAGS $CPPFLAGS" -DCMAKE_CXX_FLAGS="$CXXFLAGS $CPPFLAGS"`?

Desour:
`CFLAGS` and `CXXFLAGS` are already used to initialize `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS` (see <https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html>, <https://cmake.org/cmake/help/latest/envvar/CFLAGS.html>).
We should probably use those cmake variables, and also add the preprocessor flags (`CPPFLAGS`).

numberZero:
> `CFLAGS` and `CXXFLAGS` are already used to initialize `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS` (see https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html, https://cmake.org/cmake/help/latest/envvar/CFLAGS.html).

They are, but not `CPPFLAGS`. And specifying `CMAKE_CXX_FLAGS` explicitly on the command line overrides `CXXFLAGS` IIRC.

> We should probably use those cmake variables

No, CMake does that itself.

> and also add the preprocessor flags (`CPPFLAGS`).

If you *really* want to use that, add a cache variable (e.g. `CPP_FLAGS`), initialize it to `CPPFLAGS`, and add to compilation flags of everything (either via `add_compile_options` or non-cache `CMAKE_CXX_FLAGS`). Going without a dedicated cache variable would lead to very fragile builds.

source/Irrlicht/CMakeLists.txt
```diff
@@ -14,8 +14,9 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "^(GNU|Clang|AppleClang)$")
-	set(CMAKE_CXX_FLAGS_RELEASE "-O3")
-	set(CMAKE_CXX_FLAGS_DEBUG "-g")
+	set(CMAKE_CXX_FLAGS_RELEASE "-O3 $ENV{CPPFLAGS}")
+	set(CMAKE_CXX_FLAGS_DEBUG "-g $ENV{CPPFLAGS}")
```
numberZero, line 18: Fragile. Each time CMake runs, it rereads these environment variables.

Desour:
https://irc.minetest.net/minetest-dev/2023-05-14#i_6083236

okias:
https://github.com/minetest/minetest/pull/12799#issuecomment-1983650909

PR #136, opened 2022-10-09
Replace _IRR_OVERRIDE_ macro with override keyword

JosiahWI:
The commit also establishes a precedent of leaving off the `virtual` keyword in overrides. Although not strictly necessary, I believe this is good for readability because it makes it clear it is an override and not a pure virtual function, and it helps keep line lengths shorter.

### DIFF
--- a/include/CIndexBuffer.h
+++ b/include/CIndexBuffer.h
@@ -39,46 +39,46 @@ namespace scene
 		public:
 			core::array<T> Indices;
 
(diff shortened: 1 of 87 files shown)


JosiahWI:
I didn't notice it when I first checked, but I missed a space in my sed expression and now all the override declarations are misaligned. I'll opt to rebase everything into one commit and force push.

nerzhul (approved):

PR #138, opened 2022-10-13
win32: Use the unicode window instead of ansi window

snowyu:


source/Irrlicht/CIrrDeviceWin32.cpp
```diff
@@ -822,7 +822,7 @@ CIrrDeviceWin32::CIrrDeviceWin32(const SIrrlichtCreationParameters& params)
 		// if there is an icon, load it
 		wcex.hIcon = (HICON)LoadImage(hInstance, __TEXT("irrlicht.ico"), IMAGE_ICON, 0,0, LR_LOADFROMFILE | LR_DEFAULTSIZE);
```
sfan5, line 823: you can use `LoadImageW` here too

sfan5 (approved):
Works

sfan5:
You missed the call to `UnregisterClass` elsewhere in the file.

sfan5 (approved):

PR #139, opened 2022-10-13
#fix(win32): Can not paste the whole unicode text from clipboard on Windows

snowyu:


source/Irrlicht/COSOperator.cpp
```diff
@@ -107,19 +107,24 @@ void COSOperator::copyToClipboard(const c8 *text) const
...
-	wchar_t * buffer = 0;
-
-	HANDLE hData = GetClipboardData( CF_UNICODETEXT );
-	buffer = (wchar_t*) GlobalLock( hData );
-
-	core::wStringToMultibyte(ClipboardBuf, buffer);
+	char * buffer = 0;
 
-	GlobalUnlock( hData );
+	HANDLE hData = GetClipboardData( CF_TEXT );
```
sfan5, line 117: [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats) says this format is ANSI text, which means whatever random codepage the user has and _not_ necessarily Unicode.

what was wrong with using `CF_UNICODETEXT`?

snowyu: Only first unicode char be pasted if using CF_UNICODETEXT.

sfan5: That sounds like a conversion issue, I'll try to investigate the bug.

SmallJoker: FWIW, Wine does some polling to open the clipboard. This might be related to pending locks from other threads.

https://github.com/wine-mirror/wine/blob/master/dlls/user32/tests/clipboard.c#L38-L65

sfan5:
Fixed the issue in 56083d03e1f18955cb9a7108598c522f335e4587

PR #142, opened 2022-10-15
Add window-maximization creation-param, and isWindowMaximized()

Desour:
Needed because big centered windows don't work well.

<details>

In X11 we don't support centering the window, instead we always use pos (0,0). In SDL, the window can be (and is, because we use pos (-1,-1) in minetest) centered, but it centers into the middle of display area, not usable display area (the usable one doesn't include the panel), which is better for small windows, but completely wrong for big ones, that almost fill the usable area. And changing that (in SDL) would require repositioning the window when the window decoration is added. All in all, it would be complicated, and it's hard to decide *when* to center with respect to which area.

Fix: Keep centering the window in SDL, but store whether the window is maximised (just like `screen_[h,w]`), and maximise window on startup again accordingly.

</details>

See also: https://github.com/minetest/minetest/pull/12861

**Edit: Please don't merge yet.**

### DIFF
--- a/include/IrrCompileConfig.h
+++ b/include/IrrCompileConfig.h
@@ -45,7 +45,7 @@
 //! Example: NO_IRR_COMPILE_WITH_X11_ would disable X11
 
 //! Uncomment this line to compile with the SDL device
-//#define _IRR_COMPILE_WITH_SDL_DEVICE_
+#define _IRR_COMPILE_WITH_SDL_DEVICE_
 #ifdef NO_IRR_COMPILE_WITH_SDL_DEVICE_
 #undef _IRR_COMPILE_WITH_SDL_DEVICE_
 #endif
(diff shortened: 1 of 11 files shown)


Desour:
Rebased.

SmallJoker (approved):
Looks correct.

Desour:
(Rebased.)
Please wait with merging.
I'd like to see if it's better to use the already existing maximizeWindow function, and implement the isMaximized on other backends. (Idk why I haven't done this in the first place...)

Desour:
Added simple implementations (via `maximizeWindow()`) for the creation-params on win32 and linux backends (macOs doesn't support that function). This means that the setting now also works on these backends. But the autosaving doesn't work (same situation as with `fullscreen`) (I'd have to look up how querying the actual state works, and that's not worth it (because you'd have to review)).
(I've kept the creation-param as it makes sense to have it.)
See also the new commit.
Someone should test the win32 backend, I can't.

Desour:
(Oh, does only one review-request work? That button is weird. x))

SmallJoker (approved):
Tested with CIrrDeviceLinux + Minetest PR.
No issues found. SDL part LGTM.

sfan5 (approved):
Looks good.

PR #144, opened 2022-11-12
Fix accidental menu exit in SDL driver

ndren:
This should fix the issue described in https://github.com/minetest/minetest/issues/12941 and https://github.com/minetest/irrlicht/issues/137#issuecomment-1279770506, using the patch linked in https://github.com/minetest/minetest/issues/12941#issuecomment-1307883503. The comment there explains the caveats to this approach:
> This works, but exposes another problem. When you hit "i" to open the inventory, SDL generates a KEYDOWN followed by a TEXTINPUT. These translate into EET_KEY_INPUT_EVENT followed by EET_STRING_INPUT_EVENT, both for "i". The first one (with Char == 0) opens the inventory menu. But then the second types the letter i. This means if you open the inventory and the inventory page happens to have an edit box, it will immediately type an "i" into it.

I have not tested this myself yet (I lost my SDL build), but it should improve the problem significantly.


source/Irrlicht/CGUIEditBox.cpp
```diff
@@ -421,6 +421,19 @@ bool CGUIEditBox::processKey(const SEvent& event)
...
+		// This is needed by the SDL driver, because SDL generates both a KEYDOWN and
+		// TEXTINPUT event for the same key press. To avoid duplicating, KEYDOWN
+		// events will almost always have Char == 0.
+		if (    (event.KeyInput.Key == KEY_SPACE) ||
+			(event.KeyInput.Key >= KEY_KEY_0 && event.KeyInput.Key <= KEY_KEY_Z) ||
+			(event.KeyInput.Key >= KEY_NUMPAD0 && event.KeyInput.Key <= KEY_DIVIDE) ||
+			(event.KeyInput.Key >= KEY_OEM_1 && event.KeyInput.Key <= KEY_OEM_102))
+		{
+			return true;
+		}
```
sfan5, line 435: Sorry but this is just piling more hacks ontop of hacks, why can't `CIrrDeviceSDL.cpp` do this correctly on its own?

Desour:
It looks to me like we can't fix this in a non-hacky way without changing sdl.
Any key(-combination) can, but does not have to, make text. The proper way to find out whether it does seems to be `XLookupString` (and friends), at least that's what we're using in our X11 driver. SDL also uses it to get the text for the text event (see [here](https://github.com/libsdl-org/SDL/blob/22354b414276872f50218a9596d10fd2f4d0ad82/src/video/x11/SDL_x11events.c#L1061-L1070)), but it doesn't expose this information in the keypress events.

Desour:
AFAIK, we never call `SDL_StartTextInput`, but text input is enabled by default (according to https://github.com/libsdl-org/SDL/issues/2897). For IMEs, we should probably also call `SDL_SetTextInputRect`, handle `SDL_TEXTEDITING`, and also do everything else mentioned at <https://wiki.libsdl.org/Tutorials-TextInput>.

But the issue we're having here is that we can't distinguish keypress events that we want to use for keybinds and those that we want to ignore because they will be used to write text ("will be" because text events come after their keypress events on X11).

Maybe a better hack would be (at least on X11, and wayland) to check if the event after the key event is a text event.

paradust7:
One nice (but tricky) solution would be a keyboard text mode toggle presented by Irrlicht (similar to CursorControl) that enables and disables text input. To be enabled only when a text box has focus. When enabled, text characters only send an EET_STRING_INPUT_EVENT event (not EET_KEY_INPUT_EVENT), and vice versa when disabled. Then SDL_StartTextInput / SDL_StopTextInput can be used to suppress duplicate SDL_TEXTINPUT events.

The tricky part (maybe) is making sure the toggle is always in the correct mode.

sfan5:
Shouldn't be so tricky, we already have `acceptsIME`.

ndren:
Do you mean something like this? (Can't test myself for now.)
```
diff --git a/source/Irrlicht/CIrrDeviceSDL.cpp b/source/Irrlicht/CIrrDeviceSDL.cpp
index 12574d5..c412650 100644
--- a/source/Irrlicht/CIrrDeviceSDL.cpp
+++ b/source/Irrlicht/CIrrDeviceSDL.cpp
@@ -600,12 +600,15 @@ bool CIrrDeviceSDL::run()
 
 		case SDL_TEXTINPUT:
 			{
-				irrevent.EventType = irr::EET_STRING_INPUT_EVENT;
-				irrevent.StringInput.Str = new core::stringw();
-				irr::core::multibyteToWString(*irrevent.StringInput.Str, SDL_event.text.text);
-				postEventFromUser(irrevent);
-				delete irrevent.StringInput.Str;
-				irrevent.StringInput.Str = NULL;
+				gui::IGUIElement *elem = GUIEnvironment->getFocus();
+				if (elem && elem->acceptsIME()) {
+					irrevent.EventType = irr::EET_STRING_INPUT_EVENT;
+					irrevent.StringInput.Str = new core::stringw();
+					irr::core::multibyteToWString(*irrevent.StringInput.Str, SDL_event.text.text);
+					postEventFromUser(irrevent);
+					delete irrevent.StringInput.Str;
+					irrevent.StringInput.Str = NULL;
+				}
 			}
 			break;
 

```

source/Irrlicht/CGUIEditBox.cpp
ndren: Fixed after the force push.

ndren:
@sfan5 I've sorted it out. Now the code no longer sends Irrlicht events unnecessarily, so there is no risk of them hitting the wrong code path and having side effects (isNoUnicodeKey is back!).

Desour:
`isNoUnicodeKey` seems to work correctly, even with different keyboard layouts, as we're using keycodes and not symcodes. (Tested a bit.) I guess I worried too much.

--------

API-wise it would be better to make this depend on text input mode, not gui element focus, as sfan5 already said. (Interacting with gui stuff from the driver is not good.)

------

The enter key doesn't work in formspec fields and chat input.

ndren:
@Desour Thanks! The problem is SDL_StartTextInput and SDL_StopTextInput only affect the events that return IME Unicode, which is not the same as the events that cause the menu to exit. The events that cause the menu to exit have to, at least partially, always listen because of at least the Escape key which should work anywhere. How could this depend on text input mode instead?

ndren:
I've included KEY_RETURN and continued in #146.

Desour:
>How could this depend on text input mode instead?

Check for `SDL_IsTextInputActive()` instead of `focused_elem->acceptsIME()`.

ndren:
@Desour thanks for the suggestion! Let me know how the new code in #146 looks, it seems to have fixed the issues well: https://github.com/minetest/irrlicht/pull/146/commits/e5f9d7c8cd57a4acfed4672bd395514bddc7514b

ndren:
@paradust7 would you mind taking a look at the new PR? This one got autoclosed, but the new one looks quite a bit better in terms of being reliable and extensible.

PR #146, opened 2022-11-21
Refactor SDL input code to fix menu exit

ndren:
Continued from #144, includes KEY_RETURN.

### DIFF
--- a/source/Irrlicht/CIrrDeviceSDL.cpp
+++ b/source/Irrlicht/CIrrDeviceSDL.cpp
@@ -179,6 +179,7 @@ int CIrrDeviceSDL::findCharToPassToIrrlicht(int assumedChar, EKEY_CODE key) {
 		case KEY_UP:
 		case KEY_RIGHT:
 		case KEY_DOWN:
+		case KEY_NUMLOCK:
 			return 0;
 		default:
 			return assumedChar;
--- a/source/Irrlicht/CIrrDeviceSDL.cpp
+++ b/source/Irrlicht/CIrrDeviceSDL.cpp
@@ -8,6 +8,8 @@
 
 #include "CIrrDeviceSDL.h"
 #include "IEventReceiver.h"
+#include "IGUIElement.h"
+#include "IGUIEnvironment.h"
 #include "os.h"
 #include "CTimer.h"
 #include "irrString.h"
@@ -104,6 +106,92 @@ EM_BOOL CIrrDeviceSDL::MouseLeaveCallback(int eventType, const EmscriptenMouseEv
 	return EM_FALSE;
 }
 #endif
+
+
+bool CIrrDeviceSDL::keyIsKnownSpecial(EKEY_CODE key) const
+{
+	switch ( key )
+	{
+		// keys which are known to have safe special character interpretation
+		// could need changes over time (removals and additions!)
+		case KEY_RETURN:
+		case KEY_PAUSE:
+		case KEY_ESCAPE:
+		case KEY_PRIOR:
+		case KEY_NEXT:
+		case KEY_HOME:
+		case KEY_END:
+		case KEY_LEFT:
+		case KEY_UP:
+		case KEY_RIGHT:
+		case KEY_DOWN:
+		case KEY_TAB:
+		case KEY_PRINT:
+		case KEY_SNAPSHOT:
+		case KEY_INSERT:
+		case KEY_BACK:
+		case KEY_DELETE:
+		case KEY_HELP:
+		case KEY_APPS:
+		case KEY_SLEEP:
+		case KEY_F1:
+		case KEY_F2:
+		case KEY_F3:
+		case KEY_F4:
+		case KEY_F5:
+		case KEY_F6:
+		case KEY_F7:
+		case KEY_F8:
+		case KEY_F9:
+		case KEY_F10:
+		case KEY_F11:
+		case KEY_F12:
+		case KEY_F13:
+		case KEY_F14:
+		case KEY_F15:
+		case KEY_F16:
+		case KEY_F17:
+		case KEY_F18:
+		case KEY_F19:
+		case KEY_F20:
+		case KEY_F21:
+		case KEY_F22:
+		case KEY_F23:
+		case KEY_F24:
+		case KEY_NUMLOCK:
+		case KEY_SCROLL:
+		case KEY_LCONTROL:
+		case KEY_RCONTROL:
+			return true;
+
+		default:
+			return false;
+	}
+}
+
+int CIrrDeviceSDL::findCharToPassToIrrlicht(int assumedChar, EKEY_CODE key) {
+	switch (key) {
+		case KEY_PRIOR:
+		case KEY_NEXT:
+		case KEY_HOME:
+		case KEY_END:
+		case KEY_LEFT:
+		case KEY_UP:
+		case KEY_RIGHT:
+		case KEY_DOWN:
+			return 0;
+		default:
+			return assumedChar;
+	}
+}
+
+void CIrrDeviceSDL::resetReceiveTextInputEvents() {
+	gui::IGUIElement *elem = GUIEnvironment->getFocus();
+	if (elem && elem->acceptsIME())
+		SDL_StartTextInput();
+	else
+		SDL_StopTextInput();
+}
 
 //! constructor
 CIrrDeviceSDL::CIrrDeviceSDL(const SIrrlichtCreationParameters& param)
@@ -622,6 +710,13 @@ bool CIrrDeviceSDL::run()
 				else
 					key = (EKEY_CODE)KeyMap[idx].Win32Key;
 
+				// Make sure to only input special characters if something is in focus, as SDL_TEXTINPUT handles normal unicode already
+				// Use auto keyword to avoid typing long
+				// type definitions to get the timepoint
+				// at this instant use function now()
+				if (SDL_IsTextInputActive() && !keyIsKnownSpecial(key) && (SDL_event.key.keysym.mod & KMOD_CTRL) == 0)
+					break;
+
 #ifdef _IRR_WINDOWS_API_
 				// handle alt+f4 in Windows, because SDL seems not to
 				if ( (SDL_event.key.keysym.mod & KMOD_LALT) && key == KEY_F4)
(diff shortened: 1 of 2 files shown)


source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -680,7 +742,13 @@ bool CIrrDeviceSDL::run()
...
 			break;
 		} // end switch
-
+               gui::IGUIElement *elem = GUIEnvironment->getFocus();
+               if (elem && elem->acceptsIME())
+               {
+                       SDL_StartTextInput();
+               } else {
+                       SDL_StopTextInput();
+               }
```
Desour, line 751: Add a function in iirrdevice to set this. Gui stuff can then call the function.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -105,6 +107,67 @@ EM_BOOL CIrrDeviceSDL::MouseLeaveCallback(int eventType, const EmscriptenMouseEv
...
+		case KEY_NUMLOCK:
+		case KEY_SCROLL:
+		case KEY_LCONTROL:
+		case KEY_RCONTROL:
+		return true;
+
+	default:
+		return false;
+	}
+}
```
JosiahWI, line 169: I noticed this is pretty lengthy. Would it work to put these in an std::array and do a linear search (std::array::find)?

ndren: That is possible, but it would not make the code any shorter unless:
1. It would go way off the screen with a very long line, or
2. It would arbitrarily start newlines in some places.

Either way, it would break convention as all the other sections of code use switch cases. So I don't see that as an improvement.

ndren:
@paradust7 Please let me know if you have some feedback for this approach. Thanks!

ndren:
This PR is ready for review. I've fixed all the regressions I could find.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -105,6 +107,91 @@ EM_BOOL CIrrDeviceSDL::MouseLeaveCallback(int eventType, const EmscriptenMouseEv
...
+		case KEY_DOWN:
+			return 0;
+		default:
+			return assumedChar;
+	}
+}
+
+void CIrrDeviceSDL::resetReceiveTextInputEvents() {
+	gui::IGUIElement *elem = GUIEnvironment->getFocus();
+	if (elem && elem->acceptsIME())
```
Desour, line 189: Same suggestion as already mentioned multiple times.
Add a function in the driver that the gui code can call to enter/leave text mode. And make behaviour dependent on text mode, and do not access gui stuff from here.

ndren: So call this in minetest/minetest? Was that what you meant earlier?

ndren: So create a function "enable/disable input events" (which calls the relevant SDL toggle) and then call that in minetest/minetest when input mode changes, is that what you mean?

Desour: Yes. But note that `field[]` uses the `CGUIEditBox` from irrlicht, so not only call it from minetest.

ndren: I've condensed that part of the code down to a function call, receiveTextInputEvents, but I'm not sure where I'd need to call that. Any suggestions?

Desour: * When creating the driver (text mode is on by default according to some sdl issue).
* For cguieditbox and guieditbox in minetest in the event handler, there should be events for focus gain and lost, see e.g. here: https://github.com/minetest/irrlicht/blob/29a448de4d16ed4360e31e3e71dd79a269f52404/source/Irrlicht/CGUIEditBox.cpp#L238
* In minetest in the chat console thing when it's opened / closed.

(Btw. I should mention that you don't *have* to follow my suggestions, as I'm not a coredev. I just suggested this again because it seemed to me like you didn't quite understand me.)

ndren: The change does sound sensible, it would be good to separate out focus/unfocus from the SDL event loop. I was a bit worried that it would be quite hard to find exactly all the places in the code where focus is gained and lost. (Granted, my tool of choice is grep so it might just be that.)

FWIW, CIrrDeviceLinux.cpp (X11) runs acceptsIME directly as well.

I'll give this a few more days to see if I can find them and figure it out but if not I think I'll stick with the original code, if that's fine with you. If you decide to write one, I am happy to take up a direct patch as well.

include/IGUIEnvironment.h
```diff
@@ -411,6 +411,10 @@ class IGUIEnvironment : public virtual IReferenceCounted
 	Unless you create your own GUI elements removing themselves you won't need it.
 	\param element: Element to remove */
 	virtual void addToDeletionQueue(IGUIElement* element) = 0;
+
+	//! Enables or disables text (IME) input events.
+	/** Currently specific to SDL, though that is not necessary */
+	virtual void receiveTextInputEvents(bool receive) {};
```
sfan5, line 417: this has no implementation

sfan5:
Why are we adding a new, different method now? We have a working IME model that (as far as I can tell) SDL just needs to use, please read https://github.com/minetest/irrlicht/pull/144#issuecomment-1315073020.

Desour:
> Why are we adding a new, different method now? We have a working IME model that (as far as I can tell) SDL just needs to use

Sorry, that was an unnecessary idea by me, I was stupid and didn't read carefully enough.
The idea was that the driver could be more independent of any gui systems by offering a function that en-/disables text-input, and that key events are not fired for text keys if in text input mode.

>please read [#144 (comment)](https://github.com/minetest/irrlicht/pull/144#issuecomment-1315073020).

If I understand correctly, starting and stopping SDL text input doesn't solve the issue. It just en-/disables the extra text events. But the issue is that there are also key events that cause keybindings in minetest to be triggered, i.e. inventory key. According to <https://github.com/libsdl-org/SDL/issues/2897> (and <https://github.com/libsdl-org/SDL/issues/4739>), text input is on by default anyway.

ndren:
@sfan5 I went back to the original plan, thanks. Let me know how it looks now.
@Desour Although disabling/enabling does not do anything by itself, it does mean that the keyboard shortcut events for unicode keys (including the one for the inventory!) can be thrown away when something is in focus (so the other unicode handler deals with it), hence solving the issue.

ndren:
Side note: CI decided not to run (cancelled itself) on macos. Not sure why or how to fix.

ndren:
I squashed the commits together, hopefully now CI decides to run.

ndren:
CI builds well now, let me know if there's anything else to be done.

source/Irrlicht/CIrrDeviceSDL.h
```diff
@@ -265,6 +265,15 @@ namespace irr
 	static EM_BOOL MouseLeaveCallback(int eventType, const EmscriptenMouseEvent *mouseEvent, void *userData);
 
 #endif
+		// Check if a key is a known special character with no side effects on text boxes.
+		bool keyIsKnownSpecial(EKEY_CODE key) const;
+
+		// Return the Char that should be sent to Irrlicht for the given key (either the one passed in or 0).
+		int findCharToPassToIrrlicht(int assumedChar, EKEY_CODE key);
```
sfan5, line 272: these can be `static`

ndren:
I've had to remove const when adding static, let me know if that works.

ndren:
@sfan5 I've had to remove const when adding static, let me know if that works.

sfan5:
@Desour if you have the time, can you confirm for me whether this works?

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -105,6 +107,92 @@ EM_BOOL CIrrDeviceSDL::MouseLeaveCallback(int eventType, const EmscriptenMouseEv
...
+		case KEY_END:
+		case KEY_LEFT:
+		case KEY_UP:
+		case KEY_RIGHT:
+		case KEY_DOWN:
+			return 0;
+		default:
+			return assumedChar;
+	}
+}
```
Desour, line 186: Pressing num-lock in an edit-box produces some sort of horizontal tab for me.
This fixes it:
```cpp
	// see https://wiki.libsdl.org/SDL2/SDLKeycodeLookup
	if (assumedChar & (1<<30))
		return 0;
```

Idk if the other cases still need to return 0 because "These keys are handled differently in CGUIEditBox.cpp".
(But it doesn't seem so, according to my testing.)

(Note: The sdl keycode is a unicode char if it's printable. I'm not sure if it's appropriate to convert it to a `wchar_t`.)

Desour (commented):
Tested.
Works fine otherwise.

Desour:
> I've had to remove const when adding static, let me know if that works.

Btw. just FYI, @ndren, the const after a member function is for the implicit `this` ptr (and hence also all non-static class member variables), which doesn't exist for static member functions.

source/Irrlicht/CIrrDeviceSDL.cpp
ndren: I've added numlock as a special case, let me know if that works.

ndren:
@Desour Thanks for the clarification and testing!

source/Irrlicht/CIrrDeviceSDL.cpp
Desour: Works.
But I'd still recommend to add the check I gave above, as the keycode is not a character if that bit is set.
To emphasize what the wiki says (see link):
>Keycodes for keys with printable characters are represented by the character byte in parentheses. Keycodes without character representations are determined by their [scancode](https://wiki.libsdl.org/SDL2/SDL_Scancode) bitwise OR-ed with 1<<30 (0x40000000).

Also (see <https://github.com/libsdl-org/sdl2-compat/blob/bb66ed056b8a803677b23082438677a5676e3f22/include/SDL2/SDL_keycode.h#L37-L40>):
> Values of this type are used to represent keyboard keys using the current
> layout of the keyboard.  These values include Unicode values representing
> the unmodified character that would be generated by pressing the key, or
> an SDLK_* constant for those keys that do not generate characters.

ndren: That's a really nice way to check, thanks!

ndren:
@Desour Can you confirm this works with the new SDL check included?

Desour:
:+1: Works.
(Code is also fine, btw..)

PR #151, opened 2022-12-29
Use swap_control from MESA and EXT before SGI

x2048:
SGI does not support disabling vsync (interval == 0)

Fixes minetest/minetest#7597 on GLX drivers supporting MESA or EXT extensions for swap control


sfan5 (approved):
LGTM

PR #154, opened 2023-01-08
Avoid reskinning joints for animated meshes twice for each frame.

lhofhansl:
Needs testing.

The only point to getting the mesh in onAnimate seems to be recalculate the bounding box, which render does again.
Getting an animated mesh reskins it every time. And, worse, onAnimate is called when the mesh is not visible.


lhofhansl:
On busy scenes I gain easily 10 fps, with no loss of functionality that I could detect.

sfan5 (approved):
from looking at related code this appears to be fine

PR #156, opened 2023-02-18
SDL: Use SDL_WINDOW_FULLSCREEN_DESKTOP, and allow to maximize and fullscreen at once

Desour:
`SDL_WINDOW_FULLSCREEN` changes the screen resolution, and it stays so even after minetest is closed.

Testing instructions:
* Enable autosave screensize.
* Make the window small.
* Enable fullscreen.
* Restart minetest.
* master: Minetest is in low resolution. If you close it, everything is still in this resolution. This is bad.
* PR: Desktop resolution is used.

=> This is a bugfix for SDL.

Note that emscriptem still uses `SDL_WINDOW_FULLSCREEN`. `isFullscreen()` is only implemented in emscriptem, and emscriptem uses `isFullscreen()` for other things, so I didn't want to break anything there.
Testing for `SDL_WINDOW_FULLSCREEN_DESKTOP` in `isFullscreen()` didn't work for me btw. (X11 with openbox). The bit was never set. Hence I didn't implement it.

-----

It is now also possible to set maximized and fullscreen at once.
Sadly however, this isn't correctly stored in SDL's window flags, so autosaving maximized will disable it every time. But well, maybe someone wants to enable fullscreen and maximized and disable autosaving.

### DIFF
--- a/source/Irrlicht/CIrrDeviceSDL.cpp
+++ b/source/Irrlicht/CIrrDeviceSDL.cpp
@@ -145,12 +145,12 @@ CIrrDeviceSDL::CIrrDeviceSDL(const SIrrlichtCreationParameters& param)
 #else
 		SDL_Flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 #endif
-	} else  {
-		if (Resizable)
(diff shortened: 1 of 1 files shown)


sfan5 (approved):
lgtm

PR #161, opened 2023-02-19
Fix line endings

numberZero:
Most files were stored with CRLF line endings, while the proper way is to store LF and, for Windows users only, convert to CRLF on checkout.
Review with `git diff --ignore-space-at-eol`.

The command used was:
`sed -E -i 's/\r$//' many-wildcards-here`


sfan5:
I don't like the line endings either but I'm not convinced we want to ruin any existing diffs/PRs _now_.
I'd rather do this when the source is being imported into Minetest (which is going to lose us the commit history anyway).

SmallJoker:
:-1: breaks open PRs and complicates `git blame` unnecessarily.

numberZero:
Okay, I admit this is a sledgehammer solution. My actual problem was specific editor behavior¹, but I’ve found how to deal with it. Still,

> imported into Minetest (which is going to lose us the commit history anyway)

To my knowledge, it is possible to keep history. So unless you *decide* not to, you will need such commit, sooner or later.

> complicates `git blame`

 `git blame -w` seems to work well. (not that I knew it before yesterday)

¹ My editor apparently obeys editorconfig’s `end_of_line` on save, *even if the file had different line endings originally.* Thus touching any file with CRLF leads to a full-file diff due to CRLF→LF change. Adding `root = true` to Irrlicht’s editorconfig fixes the problem (it has no `end_of_line` setting as line ends are known mixed, but two levels up there is Minetest whose editorconfig *does* have one).

sfan5:
> To my knowledge, it is possible to keep history. So unless you _decide_ not to, you will need such commit, sooner or later.

Should be possible yeah. My intention was not to.

Issue #160, opened 2023-02-18
Make unified GL3/GLES2 driver


numberZero:
OpenGL 3 Core and OpenGL ES 2 are very similar. It should be possible to adapt `COGLES2Driver` to support both.

numberZero:
The plan is:
1. #163 
2. #167 
3. Do other stuff, in any order:
   * SDL+Android combo
   * SDL+iOS combo (if iOS is a supported platform)
   * OpenGL 3 Core-compatible driver (required for Mac)
   * WebGL-compatible driver (required for Emscripten)
   * Drop unused stuff from the unified driver
   * Merge COpenGLCore* stuff into the unified driver
4. Drop legacy devices
5. Drop legacy drivers

numberZero:
The driver is merged.

PR #167, opened 2023-02-27
Add unified OpenGL driver

numberZero:
This PR adds a unified OpenGL 3/OpenGL ES 2 driver that works on SDL.

First part of #160.
Based on #163.

On my system, works with `USE_SDL2` and `ENABLE_OPENGL3` here, `ENABLE_GLES` in Minetest, `video_driver=opengl`, and [shaders](https://github.com/minetest/irrlicht/files/10840295/Irrlicht.zip) in usual location (client/shaders/Irrlicht).

source/Irrlicht/OpenGL is basically a copy of the OGLES2 driver, with only minor changes.


numberZero:
A tricky thing is GL/GLSL version compatibility. GLSL 1.50 is supported by GL 3.2, and GL 4.2+, but is *not* guaranteed to be supported by anything in-between. And GLSL ES (both 1.00 and 3.00) is only officially supported since GL 4.3. So... dirty tricks of prepending `#version` may be needed like in `shader.cpp` in Minetest, unless actual platforms do support GLSL ES on regular GL. So, here are shaders in GLSL ES 1.00, for testing: [Irrlicht-shaders-glsl-es-1.00.zip](https://github.com/minetest/irrlicht/files/10840460/Irrlicht-shaders-glsl-es-1.00.zip).

numberZero:
Note: this does not work on core profile currently. So Mac is definitely out. But other platforms usually support compatibility profile so should work.

x2048:
What is so good in GLSL 1.50 and can we get away with a polyfill?

Do you plan to support GL2 and GL3 side by side in a single binary (I guess no.)?

numberZero:
> Do you plan to support GL 2 and GL 3 side by side in a single binary (I guess no.)?

No. And not in a single binary either. GL 3+ only.
But supporting GL 2 **ES** and GL3 in a single binary is possible.

> What is so good in GLSL 1.50

Nothing special, it’s just latest version supported by GL 3.2. The problem is, there is *no* GLSL version that is guaranteed to be supported in all GL versions since 3.2, let alone GL ES. Not a big problem actually, more a nuisance as the versions available are mostly compatible.

numberZero:
This now works with both OpenGL 3+ (4.6 on my system) and ES 2+ (3.2 on my system) on SDL2, with only minor changes to Minetest: https://github.com/minetest/minetest/commit/042d43ddbdad0c459be8e7a5fb0e39b51954ca02.
To test with GL3, set `video_driver` to `opengl3`.
To test with GLES2, set `video_driver` to `ogles2`.
One thing to notice: it uses *the exact same* video driver for both. Only context creation differs.

numberZero:
This PR is ready, but #163 has to be merged first.

sfan5:
Please rebase.

sfan5:
A couple of warnings:

<pre>
CMake Warning (dev) at source/Irrlicht/CMakeLists.txt:380 (target_sources):
  Policy CMP0076 is not set: target_sources() command converts relative paths
  to absolute.  Run "cmake --help-policy CMP0076" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

  An interface source of target "IRRVIDEOOBJ" has a relative path.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at source/Irrlicht/CMakeLists.txt:390 (target_sources):
  Policy CMP0076 is not set: target_sources() command converts relative paths
  to absolute.  Run "cmake --help-policy CMP0076" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

  An interface source of target "IRRVIDEOOBJ" has a relative path.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at source/Irrlicht/CMakeLists.txt:396 (target_sources):
  Policy CMP0076 is not set: target_sources() command converts relative paths
  to absolute.  Run "cmake --help-policy CMP0076" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

  An interface source of target "IRRVIDEOOBJ" has a relative path.
This warning is for project developers.  Use -Wno-dev to suppress it.
</pre>

Since our minimum is cmake 3.5 it looks like you'll have to use absolute paths.

sfan5:
> > Please rebase.
> 
> Merge is the Git way.

Sure but I was expecting you to resolve the conflicts.

<hr />

other feedback:

1. good job!
2. shaders should be part of this PR and committed to `media/Shaders/`, this matches the existing deployment way on Android.
3. item meshes (e.g. the hand) do not render at all, neither does the character mesh
4. when wielding a stone node it's just a solid color
5. fog does not work
6. there's some kind of depth sorting issue here: https://0x0.st/Hom8.mp4
7. ogles2 is broken now, just prints a bunch of these:
<pre>
2023-03-24 17:41:07: ERROR[Main]: Irrlicht: GLSL shader program failed to link
2023-03-24 17:41:07: ERROR[Main]: Irrlicht: error: vertex shader lacks `main'
</pre>

sfan5:
FWIW aside from the merge conflict and point 2 I'm fine with merging this early to have a basis for continuing the work.

numberZero:
> Sure but I was expecting you to resolve the conflicts.

I did that. I guess the current conflicts are new.

> shaders should be part of this PR and committed to media/Shaders/

Oops, missed that. It basically uses the same shaders that are there already but with `#version 100` prepended (which means GLSL **ES** 1.00) and under different names.

> ogles2 is broken now, just prints a bunch of these:

It requires shaders. Items 3–5 (and maybe even 6) may also be broken due to lack of shaders. Including them in Irrlicht is not enough, it will need to find them at runtime, *on all platforms.* Are Irrlicht shaders currently included in any deployment other than Android? (and for testing, you can try unpacking https://github.com/minetest/irrlicht/files/10840460/Irrlicht-shaders-glsl-es-1.00.zip into minetest/client/shaders/, it should work)

numberZero:
> shaders should be part of this PR and committed to media/Shaders/

Done.

Also, fixed line endings in all new files.

numberZero:
> Since our minimum is cmake 3.5 it looks like you'll have to use absolute paths.

Fixed with 4dda284 (relative paths are okay for private sources, and private sources should be okay here like in bfba19d67 and 10a188e96).

sfan5:
> It requires shaders. Items 3–5 (and maybe even 6) may also be broken due to lack of shaders. Including them in Irrlicht is not enough, it will need to find them at runtime, _on all platforms._

I have `client/shaders/Irrlicht` symlinked to the `media/Shaders` folder of my irrlichtmt clone, so that should work.
Unless MT forced it off I might have accidentally tested with the *MT* shaders enabled, does that matter?

> Are Irrlicht shaders currently included in any deployment other than Android? 

no

numberZero:
> I have `client/shaders/Irrlicht` symlinked to the `media/Shaders` folder of my irrlichtmt clone

:+1:
So with 28d0e06 it should work now

> Unless MT forced it off I might have accidentally tested with the _MT_ shaders enabled, does that matter?

Maybe. Without minetest/minetest#13321 MT shaders may not work well on `opengl3`, and I’m not sure whether they are disabled automatically.

numberZero:
> 7\. ogles2 is broken now, just prints a bunch of these:

Sorry but... did you test it on Emscripten? As that is the only platform Irrlicht+SDL+GLES2 could work on:

https://github.com/minetest/irrlicht/blob/ba1cd19983f9a5d6824aa0fcf0714bc102699cc9/source/Irrlicht/CIrrDeviceSDL.cpp#L486-L498

sfan5:
revised feedback:
* shader files work OOTB now
* issues 3-6 only happened with video_driver=opengl3 & enable_shaders=true, disabling shaders works fine
  * I suppose we'll need changes in MT to make all of it work
* ogles2 see below



> Sorry but... did you test it on Emscripten? In `master`, Irrlicht+SDL+GLES2 wasn’t supported on anything else.

No.
All I did was `./bin/minetest --config <(echo video_driver=ogles2;echo enable_shaders=true)` and it correctly and successfully launches a Minetest window and the about tab confirms that it's using GLES.
Now that you've pointed this out I'm wondering how this is working at all, because it shouldn't be.

sfan5:
Figured it out, you removed that and made it work in this PR: :smile: 
grafik

numberZero:
> Edit: In case we misunderstood eachother, I was trying to verify that the GLES driver still worked. Not that GLES worked on specifically SDL2 (even though I tested that).

And you’ve apparently missed a subtle detail: SDL and non-SDL builds use different GLES2 drivers in this PR. If that is *too* subtle I can add a constant for the new one... `EDT_OPENGLES2`? It will *only* work on SDL (just like `EDT_OPENGL3`), and the original `EDT_OGLES2` will remain what it is (thus won’t work on SDL at all).

numberZero:
> issues 3-6 only happened with video_driver=opengl3 & enable_shaders=true, disabling shaders works fine
> 
>     * I suppose we'll need changes in MT to make all of it work

Updated https://github.com/minetest/minetest/pull/13321

sfan5:
> And you’ve apparently missed a subtle detail: SDL and non-SDL builds use different GLES2 drivers in this PR.

I suppose you said unified so that makes sense but it's definitely too subtle for _me_.
It might also be too subtle for Minetest in case we run into things that work with ogles2-old and don't work with ogles2-new.
So not sure if we need a different `EDT_` but at the very least CMake should indicate which driver it is going to build.

Followup question: The only technical reason the unified driver only works with SDL2 is context creation, right?

numberZero:
> ogles2-old couldn't work with SDL

It doesn’t compile with SDL IIRC. That is fixable (the unified driver is based on it after all) but I’d rather not touch it unless absolutely necessary.

> The only technical reason the unified driver only works with SDL is context creation, right?

For now, it’s likely that.

> CMake should indicate which driver it is going to build.

Makes sense

> So not sure if we need a different `EDT_`

Only if you want both old and new GLES2 drivers compiled in in a single binary, and I’m not a big fan of that. Otherwise, remember that each `EDT_` corresponds to a `video_driver` setting, it would be very stupid to have `ogles2` and `ogles2new` options out of which only one may work on any given build. (also, being config option it’d be subject to the [dangers of giving a product feature the name “new”](https://devblogs.microsoft.com/oldnewthing/20221110-00/?p=107380))

> It might also be too subtle for Minetest in case we run into things that work with ogles2-old and don't work with ogles2-new.

Bugs aside, there shouldn’t be any. And for the opposite there should be better options.

sfan5:
> That is fixable (the unified driver is based on it after all) but I’d rather not touch it unless absolutely necessary.

No need, was just curious.

> Only if you want both old and new GLES2 drivers compiled in in a single binary, and I’m not a big fan of that. Otherwise, remember that each `EDT_` corresponds to a `video_driver` setting, it would be very stupid to have `ogles2` and `ogles2new` options out of which only one may work on any given build.

Those are good arguments, I agree.
If you add that piece of info at build time this PR should be good to go.

numberZero:
> If you add that piece of info at build time this PR should be good to go.

Done.

BTW as you don’t plan to save this repo history, can you merge and not rebase this? This way, follow-ups won’t need to be rebased.

sfan5:
> BTW as you don’t plan to save this repo history, can you merge and not rebase this? This way, follow-ups won’t need to be rebased.

The idea was to not squash it yes.
I was going to ask whether you wanted to squash a few commits to minimize redundant changes (e.g. the lines endings commit) but that's unneeded extra work for you. I'll merge it as-is.

PR #170, opened 2023-02-28
Don't draw tooltips

Desour:
For https://github.com/minetest/minetest/pull/13264.

### DIFF
--- a/source/Irrlicht/CGUIEnvironment.cpp
+++ b/source/Irrlicht/CGUIEnvironment.cpp
@@ -387,6 +387,8 @@ bool CGUIEnvironment::OnEvent(const SEvent& event)
 //
 void CGUIEnvironment::OnPostRender( u32 time )
 {
+	// Minetest draws its tooltips itself
+#if 0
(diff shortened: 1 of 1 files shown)


x2048:
Why not remove the code altogether if we are not using it?

Desour:
I'd then also have to remove plenty of other stuff, like ToolTip in IGUIElement, or the stuff in the skin. It's likely I'd forget something. And there'd be an API break, which would make merging harder I guess.
Also, maybe someone considers at some point that this style of ToolTips is better (the current implementation in GUIFormSpecMenu is kinda ugly), so I didn't want to remove it yet.

sfan5:
> And there'd be an API break, which would make merging harder I guess.

FYI this doesn't matter at all

SmallJoker:
Whereas the API break is no concern, I think that offloading some code to remove clutter from GUIFormSpecMenu would make sense. If there is a reasonable way to integrate Minetest's custom tooltip code into the generic Irrlicht tooltips it could help to make the code cleaner.

PR #177, opened 2023-03-13
Remove unused dependencies from SceneManager

numberZero:
Ready for Review.

Note: SceneManager::getMesh is only used in Client::getMesh and only with a RAM file.


numberZero:
oops... I've checked Minetest and all but not tests...

PR #179, opened 2023-03-20
Cherry-pick commits from Irrlicht 

sfan5:


SmallJoker:
Might it be possible to properly attribute the original author in the commit or can this information not be carried over from SVN to Git?

sfan5:
The original author is attributed, Github just doesn't show it.

PR #180, opened 2023-03-24
Use non-static member vars for SDL clipboard / primary selection buffers

Desour:
Fixes <https://github.com/minetest/irrlicht/pull/132#discussion_r1147855997>.
(I don't remember why I used static vars in the first place...)

(Also removed the null-checks before `SDL_free`, as they're not required.)

### DIFF
--- a/source/Irrlicht/COSOperator.h
+++ b/source/Irrlicht/COSOperator.h
@@ -23,7 +23,7 @@ public:
 #endif
 	COSOperator(const core::stringc& osversion);
 
-	~COSOperator() override;
+	~COSOperator();
 
 	COSOperator(const COSOperator &) = delete;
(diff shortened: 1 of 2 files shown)


source/Irrlicht/COSOperator.h
```diff
@@ -23,6 +23,11 @@ class COSOperator : public IOSOperator
 #endif
 	COSOperator(const core::stringc& osversion);
 
+	~COSOperator() override;
```
sfan5, line 26:
```suggestion
	~COSOperator();
```

Desour: Why no override?

sfan5: Not necessary.

PR #181, opened 2023-03-25
[NOSQUASH] Move stuff from renderingengine

Desour:
Moves the platform dependent stuff from renderingengine.cpp. Also implements the stuff for SDL (2nd commit).
This fixes the remaining point in #137.

~~I don't know how to implement `getDisplayDensity` properly. `SDL_GetDisplayDPI` doesn't seem to be what we need. **Please help.**~~

FYI, this doesn't break anything if merged without https://github.com/minetest/minetest/pull/13348. (The X11 setup stuff is done twice, but that doesn't seem to hurt.)

(The X11 class name and the windows icon thing are hardcoded to minetest's values. I don't think passing the values to irrlicht is worth doing, as we're planning to drop non-SDL backends anyway.)

It would be nice if someone could test this with Windows.

### DIFF
--- a/include/IImage.h
+++ b/include/IImage.h
@@ -329,6 +329,12 @@ public:
 	//! Sets a pixel
 	virtual void setPixel(u32 x, u32 y, const SColor &color, bool blend = false ) = 0;
 
+	//! Copies this surface into another, if it has the exact same size and format.
+	/**	NOTE: mipmaps are ignored
+	\return True if it was copied, false otherwise.
+	*/
+	virtual bool copyToNoScaling(void *target, u32 width, u32 height, ECOLOR_FORMAT format=ECF_A8R8G8B8, u32 pitch=0) const = 0;
+
 	//! Copies the image into the target, scaling the image to fit
 	/**	NOTE: mipmaps are ignored */
 	virtual void copyToScaling(void* target, u32 width, u32 height, ECOLOR_FORMAT format=ECF_A8R8G8B8, u32 pitch=0) =0;
--- a/source/Irrlicht/CIrrDeviceLinux.cpp
+++ b/source/Irrlicht/CIrrDeviceLinux.cpp
@@ -24,6 +24,8 @@
 #include "SIrrCreationParameters.h"
 #include "SExposedVideoData.h"
 #include "IGUISpriteBank.h"
+#include "IImageLoader.h"
+#include "IFileSystem.h"
 #include <X11/XKBlib.h>
 #include <X11/Xatom.h>
 
@@ -171,6 +173,8 @@ CIrrDeviceLinux::CIrrDeviceLinux(const SIrrlichtCreationParameters& param)
 
 	if (param.WindowMaximized)
 		maximizeWindow();
+
+	setupTopLevelXorgWindow();
 }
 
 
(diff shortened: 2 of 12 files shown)


numberZero:
> And I don't know much about dpi stuff.

That stuff is way more complicated than it needs to be. In a nutshell,

* `DisplayHeightMM` returns `DisplayHeight`/3.78 on recent X.Org versions
* On X11 and Wayland, `SDL_GetDisplayDPI` returns actual screen unit to inch ratio (assuming no bad configuration)
* On X11, screen unit is exactly pixel (even if scaling is used: X.Org just scales the whole framebuffer)
* On Wayland, if you don’t set `SDL_WINDOW_ALLOW_HIGHDPI` the behavior is essentially the same, application-wise
* On Wayland, if you do set `SDL_WINDOW_ALLOW_HIGHDPI`, screen unit may not be the same as pixel; you can query window size in screen units with `SDL_GetWindowSize` and the same size in real pixels with `SDL_GL_GetDrawableSize`

I don’t know how does that work on other platforms.

Also, there are things like “font DPI,” “UI DPI,” etc. which aren’t DPI at all but rather, scale factors.

> Should we maybe just assume a logical density of 96 screen coordinates per inch

That is what the current code does on X11 effectively. Here is the result:
image

> I'm also not sure what `RenderingEngine::getDisplayDensity()` is actually supposed to return.

UI&font scaling factor, I suppose. That should be a user preference and not a function of DPI, but getting it is highly platform-dependent (e.g. Qt and GTK have different ways for that) so simply returning actual_dpi/96 is likely a good idea.

Desour:
Thanks for the explanations!

To get the same results as we get on non-SDL backends, I've decided to take window size / drawable size, as this is the same as in the X11 backend, according to:
 
> `DisplayHeightMM` returns `DisplayHeight/3.78` on recent X.Org versions

----

> That should be a user preference and not a function of DPI, but getting it is highly platform-dependent (e.g. Qt and GTK have different ways for that) so simply returning actual_dpi/96 is likely a good idea.

Minetest also has its `display_density_factor` setting. Maybe we could in the future add special values for this setting (i.e. 0, -1, or -2) to use the physical dpi or SDL3's `display_scale`.
For now I've decided against `SDL_GetDisplayDPI`, because it would scale differently on different IrrDevice backends, and because it will be removed in SDL3.

----

(Also rebased.)

---

This is ready for review now.

Someone should test with a windows device.

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -908,6 +933,83 @@ void CIrrDeviceSDL::setWindowCaption(const wchar_t* text)
...
+	}
+
+	io::IReadFile *icon_f = getFileSystem()->createAndOpenFile(path.c_str());
+
+	if (!icon_f) {
+		os::Printer::log(("Could not load icon file '" + path + "'").c_str(), ELL_WARNING);
+		return false;
+	}
+
+	video::IImage *img = image_loader->loadImage(icon_f);
```
numberZero, line 964: Why not `VideoDriver->createImageFromFile(path)`?

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -908,6 +933,83 @@ void CIrrDeviceSDL::setWindowCaption(const wchar_t* text)
...
+		for (u32 x = 0; x < width; x++) {
+			video::SColor col = img->getPixel(x, y);
+			u32 pixel_val = 0;
+			pixel_val |= (u32)col.getAlpha() << 24;
+			pixel_val |= (u32)col.getRed() << 16;
+			pixel_val |= (u32)col.getGreen() << 8;
+			pixel_val |= (u32)col.getBlue();
+			row[x] = pixel_val;
+		}
+	}
```
numberZero, line 998: Can’t you convert the image to ARGB with e.g. `IImage::copyTo`? Or at least use `col.color`, it is in the same format as `pixel_val` IIUC.

source/Irrlicht/CIrrDeviceSDL.cpp
Desour: I've adapted the code from the X11 thing we had in renderingengine.cpp. I didn't know this function exists.

Using `VideoDriver->createImageFromFile(path)` now, thanks!

numberZero:
Tested.

#### Window icon
X11: still works on DeviceLinux, and with minetest/minetest#13348 it now also works on DeviceSDL.
Wayland: no changes; SDL apparently doesn’t support window icon yet, so no luck here.

#### Display density
X11: no changes; that’s expected.
Wayland: no changes either. I guess it’s missing `SDL_WINDOW_ALLOW_HIGHDPI`.

source/Irrlicht/CIrrDeviceSDL.cpp
Desour: Same as above.

(But we still have to manually copy for X11, because it's an array of longs.)

Desour:
>I guess it’s missing `SDL_WINDOW_ALLOW_HIGHDPI`.

This should be done in another PR. I couldn't even test it.

source/Irrlicht/CIrrDeviceLinux.cpp
```diff
@@ -1178,6 +1248,59 @@ void CIrrDeviceLinux::setWindowCaption(const wchar_t* text)
 }
 
 
+//! Sets the window icon.
+bool CIrrDeviceLinux::setWindowIcon(const std::string &path)
```
sfan5, line 1252: should take an `IImage` IMO

Desour: Done.

Note though that win32 will now also require the (unused by win32) icon image file to work.

include/IImage.h
```diff
@@ -329,6 +329,12 @@ class IImage : public virtual IReferenceCounted
 	//! Sets a pixel
 	virtual void setPixel(u32 x, u32 y, const SColor &color, bool blend = false ) = 0;
 
+	//! Copies this surface into another, if it fits.
```
numberZero, line 332: The comment states “if it fits” but the implementation is “if it has the exact same size and format”

numberZero:
btw here is my untested version of `copyToNoScaling`
```c++
bool CImage::copyToNoScaling(void *target, ECOLOR_FORMAT format, u32 pitch) const
{
	if (IImage::isCompressedFormat(Format))
	{
		os::Printer::log("IImage::copyToNoScaling method doesn't work with compressed images.", ELL_WARNING);
		return false;
	}

	if (!target || !Size.Width || !Size.Height)
		return false;

	const u32 bpp=getBitsPerPixelFromFormat(format)/8;
	if (0==pitch)
		pitch = Size.Width*bpp;

	if (Format == format && pitch == Pitch) {
		memcpy(target, Data, (size_t)Size.Height*pitch);
		return true;
	}

	if (!CColorConverter::canConvertFormat(Format, format))
		return false;

	u8* tgtpos = (u8*) target;
	const u8* srcpos = Data;
	for (u32 y = 0; y < Size.Height; ++y) {
		CColorConverter::convert_viaFormat(srcpos, Format, Size.Width, tgtpos, format);
		tgtpos += pitch;
		srcpos += Pitch;
	}
	return true;
}
```

Desour:
Fixed the comment.
I'd like to keep the copyToNoScaling implementation simple for now (it's curretnly just copied from copyToScaling).
The image class will probably be reworked at some point anyways.

source/Irrlicht/CIrrDeviceLinux.cpp
sfan5: As of now that doesn't work, the file is not installed on Windows.
(assuming you didn't change that in the relevant PR)

Desour: (Windows installations would also need the file for the SDL driver.)

>assuming you didn't change that in the relevant PR

I didn't. Idk how to fix this, and I couldn't test.

source/Irrlicht/CIrrDeviceLinux.cpp
```diff
@@ -282,6 +286,72 @@ bool CIrrDeviceLinux::switchToFullscreen()
...
+	if (CreationParams.DriverType == video::EDT_NULL)
+		return; // no display and window
+
+	os::Printer::log("Configuring X11-specific top level window properties", ELL_DEBUG);
+
+	// Set application name and class hints. For now name and class are the same.
+	// Note: SDL uses the executable name here (i.e. "minetest").
+	XClassHint *classhint = XAllocClassHint();
+	classhint->res_name = const_cast<char *>("Minetest");
+	classhint->res_class = const_cast<char *>("Minetest");
```
SmallJoker, line 301: Technically speaking, those names do not belong to the framework. However, IrrlichtMt is basically modified to fit to Minetest's needs.

Thing is - I would like to use IrrlichtMt over stock Irrlicht for an unrelated project to benefit from the well-maintained code. However, if keeping it generalized is too much effort, I would understand that decision as well.

SmallJoker:
Tested on Linux/X11 (the classic) with the Minetest PR and could not find any issues. Also based on the buildbot results and numberZero's testing I will give an approval after clarifying the comment above.

source/Irrlicht/CIrrDeviceLinux.cpp
Desour: I have strong doubts about the usefulness of using irrlichtmt for another application, as there will always be annoying breaking changes for you.

But if you do still want to do this, you can also just choose the SDL driver, it takes these names from the executable's filename.

=> It's not worth doing imo, sorry.

SmallJoker (approved):
Works.

sfan5 (approved):
LGTM

PR #185, opened 2023-04-05
Use a buffer for quads indices

numberZero:
Follow-up to #167

Mandatory for core profile, nice to have otherwise.
Note: it may be better to limit quad count to something smaller, e.g. to fit it in `GL_MAX_ELEMENTS_VERTICES` and `GL_MAX_ELEMENTS_INDICES`
Note: it might also be better to use `glDrawRangeElements`. Or maybe instead. **UPD: Done**


numberZero:
Closed/reopened to force GitHub to recompute the diff after master branch update, as per https://github.com/orgs/community/discussions/51948#discussioncomment-5536787.

PR #187, opened 2023-04-07
Fix compile warning when appending UTF-8/ASCII strings to ustring16

### LINKED ISSUE #166: -Wtype-limits warnings in irrUString.h
I'm getting -Wtype-limits warnings in these lines:
https://github.com/minetest/irrlicht/blob/5527b9f373ef4267a2b44f816be07ddab48a289d/include/irrUString.h#L1373
https://github.com/minetest/irrlicht/blob/5527b9f373ef4267a2b44f816be07ddab48a289d/include/irrUString.h#L1474

If you cast a signed char (to compare), it's some negative number, and the comparison is always false.

Other lines may also be affected, i.e. wrong shifting.

I guess #33 introduced this. (cc @numberZero @sfan5)

Maybe using u8 instead of char would be better. But idk, hence this issue.

bengardner:
String constants are (const char *), but UTF-8 strings must be treated as an array of bytes when fiddling with the bits.

The following comparison fails without this change, as uchar8_t is a signed char, which cannot be 0xC0:

   const uchar8_t* c2 = other;
   ...
   else if (c2[l] == 0xC0 || c2[l] == 0xC1)
   ...

This patch adds the appropriate casts to fix the compile warning.

Fixes #166

### DIFF
--- a/include/irrUString.h
+++ b/include/irrUString.h
@@ -1323,7 +1323,7 @@ public:
 
 		// Determine if the string is long enough for a BOM.
 		u32 len = 0;
-		const uchar8_t* p = other;
+		const u8* p = reinterpret_cast<const u8*>(other);
 		do
 		{
 			++len;
@@ -1338,10 +1338,10 @@ public:
 		}
 
 		// If a BOM was found, don't include it in the string.
-		const uchar8_t* c2 = other;
+		const u8* c2 = reinterpret_cast<const u8*>(other);
 		if (c_bom != unicode::EUTFE_NONE)
 		{
-			c2 = other + unicode::BOM_UTF8_LEN;
+			c2 += unicode::BOM_UTF8_LEN;
 			length -= unicode::BOM_UTF8_LEN;
 		}
 


include/irrUString.h
```diff
@@ -1323,7 +1323,7 @@ class ustring16
 
 		// Determine if the string is long enough for a BOM.
 		u32 len = 0;
-		const uchar8_t* p = other;
+		const u8* p = (const u8*)other;
```
sfan5, line 1326: should be
```cpp
		const u8* p = reinterpret_cast<const u8*>(other);
```

bengardner: Yes, using reinterpret_cast also works.

The source file uses a mix of C-style casts and C++ reinterpret_cast, so I don't see that there should be a preference.

include/irrUString.h
```diff
@@ -1338,7 +1338,7 @@ class ustring16
 		}
 
 		// If a BOM was found, don't include it in the string.
-		const uchar8_t* c2 = other;
+		const u8* c2 = (const u8*)other;
 		if (c_bom != unicode::EUTFE_NONE)
 		{
 			c2 = other + unicode::BOM_UTF8_LEN;
```
bengardner, line 1344: "other" and "c2" have different types here.

include/irrUString.h
bengardner: I pushed a version using reinterpret_cast. Tested with GCC 9.4.0 on Linux.

PR #188, opened 2023-04-08
Drop obsolete loadImages and createImagesFromFile

numberZero:
They were to load several images from a single file. But IrrlichtMt doesn’t support such image formats.

Minetest doesn’t use these functions. It does use singular variants, however.

Note: review with “hide whitespace.“


Desour (approved):
Haven't tested. But code looks fine. :+1:

PR #193, opened 2023-04-18
OpenGL3: new version&feature checks

numberZero:
1. Save OpenGL spec and version in the driver (`initVersion`)
2. Fill feature support information accordingly (`initFeatures`)
3. Restore support for DrawBuffers as MRT is now properly reported as supported.

Follow-ups:
* #194 (may be considered an alternative)
* #195 
* #196

Note: [ce29446](https://github.com/minetest/irrlicht/pull/193/commits/ce29446a42866781bae447ce815af3a7cfb5c09e) is reverted in #194 as it uses a different approach.

PR #194, opened 2023-04-18
OpenGL3: New extension listing system

numberZero:
~~Based on #193. May be considered an alternative.~~ UPD: #193 merged
~~[Diff against #193](https://github.com/numberZero/irrlicht/pull/26/files).~~ UPD: obsolete due to rebase

Current extension handling is relying on the hard-coded possible extension list. Runtime checks check extension availability in this list.

The new code lists all extensions actually available, and at initialization time derives which features to enable (for now, only `AnisotropicFilterSupported` actually replaces an extension check).


sfan5 (approved):
Looks sensible

PR #197, opened 2023-04-23
Android CI

numberZero:
Add Android CI. Mostly a copy from [minetest_android_deps](https://github.com/minetest/minetest_android_deps) where Irrlicht was added before forking.

An alternative would be to use [build.sh](https://github.com/minetest/minetest_android_deps/blob/master/build.sh) directly from that repo. But keeping the build script separate from own code makes little sense IMO. A better approach could be to only import `_setup_toolchain` to ensure compatibility but that requires some changes in the deps repo.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -208,10 +208,14 @@ jobs:
           path: artifact/
 
   android:
+    name: Android ${{ matrix.arch }}
     runs-on: ubuntu-20.04
     env:
        ndk_version: "r25c"
-       ANDROID_NDK: ${{ env.HOME }}/android-ndk
+       ANDROID_NDK: ${{ github.workspace }}/android-ndk
+    strategy:
+      matrix:
+        arch: [armeabi-v7a, arm64-v8a, x86, x86_64]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


numberZero:
Note: the CI passes on its own commit, [f050847](https://github.com/minetest/irrlicht/pull/197/commits/f0508477d68ef796e0ec727032cac3ba98a0eea2): https://github.com/numberZero/irrlicht/actions/runs/4778533986/jobs/8495012548. The failure is due to later changes in this repo.

numberZero:
Now that Android fixes are in `master`, this simply works. Didn’t test whether the Irrlicht the CI builds works too, though.

sfan5:
Maybe we can just clone https://github.com/minetest/minetest_android_deps and run the build from there? I'm not too happy about so much duplication.

numberZero:
> Maybe we can just clone https://github.com/minetest/minetest_android_deps and run the build from there? I'm not too happy about so much duplication.

JPEG and PNG don’t need to be built here at all anymore as they’re available in the deps since https://github.com/minetest/minetest_android_deps/pull/4. But for Irrlicht itself, it’d be rather awkward to keep the build script in a separate repository. For the environment setup it’s possible but copy it or import it, it will be fragile either way.

sfan5:
I decided my opinion is wrong.

PR #184, opened 2023-04-05
Cleanup fixed materials

numberZero:
Follow-up to #167

PR #203, opened 2023-06-14
Remove unused fixed function materials

grorp:
This PR removes the 13 fixed function Irrlicht materials that Minetest doesn't use. It is inspired by #184, which removed the implementation of these materials in the new unified OpenGL driver.

Tested on ...

- Android with `video_driver = ogles1`
- Android with `video_driver = ogles2` + `enable_shaders = false`
- Linux with `video_driver = opengl` + `enable_shaders = false`

This is related to the roadmap goal "Cleanup" -> "Drop other features we don't use".


sfan5 (approved):
Works, thanks.

numberZero (approved):

PR #204, opened 2023-06-23
Refactor parts of the material system and set min filter to bilinear for 2D graphics

grorp:
This is the Irrlicht side of minetest/minetest#13622.

This PR fixes minetest/minetest#13621. Because it exposes all OpenGL texture filtering options to Minetest, it also makes minetest/minetest#13108 trivial to fix.

### Part 1: Refactor parts of the material system

The first 6 commits of this PR are the "Refactor parts of the Irrlicht material system" part.

Before this PR:

- `SMaterialLayer` has two boolean fields, `BilinearFilter` and `TrilinearFilter`, which select both the min and the mag filter
- you set them via `SMaterial::setFlag`, which automatically sets them on all texture layers for you

After this PR:

- `SMaterialLayer` has two enum fields, `MinFilter` and `MagFilter`, which select the min and the mag filter
- ❌ setting them via `SMaterial::setFlag`? impossible because they are no booleans
- ❌ manually setting them on all texture layers? would cause code duplication
- ✅ you set them in a lambda passed to the new `SMaterial::forEachTexture` function

However, setting a per-texture material flag on all texture layers was the only real use of `SMaterial::setFlag`. All other uses could trivially be replaced by something like `material.<name of the flag> = <value>`, i.e. `setFlag` was a simple setter function. Therefore, I replaced all of its uses and removed it.

There also was `ISceneNode::setMaterialFlag`, which set a material flag on all materials of a scene node for you. Because I needed a replacement for that as well, I introduced `ISceneNode::forEachMaterial`.

One thing I like about the new API is that it makes it explicit at what layer you are configuring something right now. Look at this code using the old API:

```c++
// sets the type of all materials of this scene node
m_spritenode->setMaterialType(video::EMT_TRANSPARENT_ALPHA_CHANNEL);
// sets a flag on all materials of this scene node
m_spritenode->setMaterialFlag(video::EMF_LIGHTING, false);
// sets the first texture of all materials of this scene node
m_spritenode->setMaterialTexture(0,
		env->getGameDef()->tsrc()->getTextureForMesh("smoke_puff.png"));
// sets a flag on all textures of all materials of this scene node
m_spritenode->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
```

Without the comments, would you have guessed what each function call does? This is the equivalent code with the new API:

```c++
video::ITexture *smoke = env->getGameDef()->tsrc()->getTextureForMesh("smoke_puff.png");

m_spritenode->forEachMaterial([smoke] (video::SMaterial &mat) {
	mat.MaterialType = video::EMT_TRANSPARENT_ALPHA_CHANNEL;
	mat.Lighting = false;
	mat.setTexture(0, smoke);
	mat.forEachTexture([] (video::SMaterialLayer &tex) {
		tex.BilinearFilter = false;
	});
});
```

The new API makes it obvious that you should actually be doing this:

```c++
video::ITexture *smoke = env->getGameDef()->tsrc()->getTextureForMesh("smoke_puff.png");

m_spritenode->forEachMaterial([smoke] (video::SMaterial &mat) {
	mat.MaterialType = video::EMT_TRANSPARENT_ALPHA_CHANNEL;
	mat.Lighting = false;
	mat.setTexture(0, smoke);
	mat.TextureLayers[0].BilinearFilter = false;
});
```

(The two "new API" examples don't take the switch from `BilinearFilter` and `TrilinearFilter` to `MinFilter` and `MagFilter` into account.)

### Part 2: Set min filter to bilinear for 2D

The last commit of this PR is the "Set min filter to bilinear for 2D graphics" part.

This commit enables bilinear filtering for downscaling textures in Irrlicht's 2D renderer (which is used by Minetest's GUIs). Essentially, it just sets `GL_TEXTURE_MIN_FILTER` to `GL_LINEAR` for 2D rendering.

## To do

This PR is a Ready for Review.

## How to test

Use Minetest on all platforms and with all Irrlicht video drivers and verify that everything looks the same as before, with one exception: Images rendered with a lower resolution than their original resolution in GUIs look better. The difference is very visible with minetest/minetest#13606, for example.

## Additional context

Minetest cannot upscale GUI images with bilinear filtering because that would blur pixel art, but we can _downscale_ images with bilinear filtering. This makes downscaled images look much better. Compare these screenshots:

\[1\] before this PR, \[2\] after this PR

screenshot before screenshot after

### DIFF
--- a/include/EMaterialProps.h
+++ b/include/EMaterialProps.h
@@ -76,7 +76,7 @@ namespace video
 		EMP_POLYGON_OFFSET = 0x80000,
 
 		//! Corresponds to SMaterial::BlendFactor.
-		EMP_BLEND_FACTOR = 0x160000
+		EMP_BLEND_FACTOR = 0x100000,
 	};
 
 } // end namespace video
--- a/include/EMaterialProps.h
+++ b/include/EMaterialProps.h
@@ -2,86 +2,86 @@
 // This file is part of the "Irrlicht Engine".
 // For conditions of distribution and use, see copyright notice in irrlicht.h
 
-#ifndef __E_MATERIAL_FLAGS_H_INCLUDED__
-#define __E_MATERIAL_FLAGS_H_INCLUDED__
+#ifndef __E_MATERIAL_PROPS_H_INCLUDED__
+#define __E_MATERIAL_PROPS_H_INCLUDED__
 
 namespace irr
 {
 namespace video
 {
 
-	//! Material flags
-	enum E_MATERIAL_FLAG
+	//! Material properties
+	enum E_MATERIAL_PROP
 	{
 		//! Corresponds to SMaterial::Wireframe.
-		EMF_WIREFRAME = 0x1,
+		EMP_WIREFRAME = 0x1,
 
 		//! Corresponds to SMaterial::PointCloud.
-		EMF_POINTCLOUD = 0x2,
+		EMP_POINTCLOUD = 0x2,
 
 		//! Corresponds to SMaterial::GouraudShading.
-		EMF_GOURAUD_SHADING = 0x4,
+		EMP_GOURAUD_SHADING = 0x4,
 
 		//! Corresponds to SMaterial::Lighting.
-		EMF_LIGHTING = 0x8,
+		EMP_LIGHTING = 0x8,
 
 		//! Corresponds to SMaterial::ZBuffer.
-		EMF_ZBUFFER = 0x10,
+		EMP_ZBUFFER = 0x10,
 
 		//! Corresponds to SMaterial::ZWriteEnable.
-		EMF_ZWRITE_ENABLE = 0x20,
+		EMP_ZWRITE_ENABLE = 0x20,
 
 		//! Corresponds to SMaterial::BackfaceCulling.
-		EMF_BACK_FACE_CULLING = 0x40,
+		EMP_BACK_FACE_CULLING = 0x40,
 
 		//! Corresponds to SMaterial::FrontfaceCulling.
-		EMF_FRONT_FACE_CULLING = 0x80,
+		EMP_FRONT_FACE_CULLING = 0x80,
 
 		//! Corresponds to SMaterialLayer::MinFilter.
-		EMF_MIN_FILTER = 0x100,
+		EMP_MIN_FILTER = 0x100,
 
 		//! Corresponds to SMaterialLayer::MagFilter.
-		EMF_MAG_FILTER = 0x200,
+		EMP_MAG_FILTER = 0x200,
 
 		//! Corresponds to SMaterialLayer::AnisotropicFilter.
-		EMF_ANISOTROPIC_FILTER = 0x400,
+		EMP_ANISOTROPIC_FILTER = 0x400,
 
 		//! Corresponds to SMaterial::FogEnable.
-		EMF_FOG_ENABLE = 0x800,
+		EMP_FOG_ENABLE = 0x800,
 
 		//! Corresponds to SMaterial::NormalizeNormals.
-		EMF_NORMALIZE_NORMALS = 0x1000,
+		EMP_NORMALIZE_NORMALS = 0x1000,
 
 		//! Corresponds to SMaterialLayer::TextureWrapU, TextureWrapV and
 		//! TextureWrapW.
-		EMF_TEXTURE_WRAP = 0x2000,
+		EMP_TEXTURE_WRAP = 0x2000,
 
 		//! Corresponds to SMaterial::AntiAliasing.
-		EMF_ANTI_ALIASING = 0x4000,
+		EMP_ANTI_ALIASING = 0x4000,
 
 		//! Corresponds to SMaterial::ColorMask.
-		EMF_COLOR_MASK = 0x8000,
+		EMP_COLOR_MASK = 0x8000,
 
 		//! Corresponds to SMaterial::ColorMaterial.
-		EMF_COLOR_MATERIAL = 0x10000,
+		EMP_COLOR_MATERIAL = 0x10000,
 
 		//! Corresponds to SMaterial::UseMipMaps.
-		EMF_USE_MIP_MAPS = 0x20000,
+		EMP_USE_MIP_MAPS = 0x20000,
 
 		//! Corresponds to SMaterial::BlendOperation.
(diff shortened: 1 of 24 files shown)


include/SMaterialLayer.h
```diff
@@ -43,14 +43,38 @@ namespace video
...
+	};
+
+	//! Texture magnification filter.
+	/** Used when scaling textures up. */
+	enum E_TEXTURE_MAG_FILTER {
+		//! Nearest-neighbor interpolation.
+		ETMAGF_NEAREST = 0,
+		//! Linear interpolation.
+		ETMAGF_BILINEAR,
+	};
```
numberZero, line 68: Maybe use actual OpenGL tokens?

include/SMaterialLayer.h
```diff
@@ -43,14 +43,38 @@ namespace video
...
+		//! Nearest-neighbor interpolation.
+		ETMINF_NEAREST = 0,
+		//! Linear interpolation.
+		ETMINF_BILINEAR,
+		//! Linear interpolation across mipmaps.
+		/** Is equivalent to ETMINF_BILINEAR if mipmaps are disabled.
+		Only available as a minification filter since mipmaps are only used
+		when scaling down. */
+		ETMINF_TRILINEAR,
+	};
```
numberZero, line 59: OpenGL has more options. Basically, you can set intra-mip-level and inter-mip-level interpolation independently.

include/EMaterialProps.h
```diff
@@ -0,0 +1,87 @@
...
+
+		//! Corresponds to SMaterial::BlendOperation.
+		EMP_BLEND_OPERATION = 0x40000,
+
+		//! Corresponds to SMaterial::PolygonOffsetFactor, PolygonOffsetDirection,
+		//! PolygonOffsetDepthBias and PolygonOffsetSlopeScale.
+		EMP_POLYGON_OFFSET = 0x80000,
+
+		//! Corresponds to SMaterial::BlendFactor.
+		EMP_BLEND_FACTOR = 0x160000
```
numberZero, line 79: Looks like a bug, 2⋅0x80000≠0x160000. `EMP_USE_MIP_MAPS` shouldn’t be included here. (I see that it was this way in the original code, i.e. it was missed earlier)

include/SMaterialLayer.h
grorp: I don't think we can use the actual `GL_…` enum values if we want proper type checking, but I'll at least give them the same names as the OpenGL enum values.

include/SMaterialLayer.h
grorp: All options provided by OpenGL are now exposed.

include/SMaterialLayer.h
grorp: The `ETMINF_…` and `ETMAGF_…` enum values now have the same names as the `GL_…` enum values.

include/EMaterialProps.h
grorp: Changed the value of `EMP_BLEND_FACTOR` to `0x100000`.
Why do you want to remove `EMP_USE_MIP_MAPS`?

grorp:
Thanks for reviewing.

include/EMaterialProps.h
numberZero: Oops, sorry. I meant that 0x160000 includes 0x20000 (EMP_USE_MIP_MAPS) but EMP_BLEND_FACTOR clearly shouldn’t include it.

grorp: Ah, okay :)

include/SMaterialLayer.h
Desour: OpenGL has 6 different modes: https://www.khronos.org/opengl/wiki/Sampler_Object#Filtering

Desour (approved):
Shouldn't there be a `IMesh::forEachMeshBuffer`?

Code looks fine. Tested, works.

grorp:
Thanks for reviewing.

> Shouldn't there be a IMesh::forEachMeshBuffer?

It would probably make sense to have that method, but as the old `IMesh::setMaterialFlag` wasn't used anywhere, I'd prefer not to add it in this PR.

sfan5:
Before I merge this can you drop the "(IrrlichtMt side)" from the commit messages?

include/EMaterialProps.h
```diff
@@ -0,0 +1,87 @@
+// Copyright (C) 2002-2012 Nikolaus Gebhardt
+// This file is part of the "Irrlicht Engine".
+// For conditions of distribution and use, see copyright notice in irrlicht.h
+
+#ifndef __E_MATERIAL_PROPS_H_INCLUDED__
+#define __E_MATERIAL_PROPS_H_INCLUDED__
```
sfan5, line 6: fyi there's no harm in changing enum values as you please
(unless you renamed it for another reason)

source/Irrlicht/CNullDriver.cpp
```diff
@@ -97,19 +97,21 @@ CNullDriver::CNullDriver(io::IFileSystem* io, const core::dimension2d<u32>& scre
...
-		InitMaterial2D.TextureLayer[i].TextureWrapV=video::ETC_REPEAT;
-		InitMaterial2D.TextureLayer[i].TextureWrapW = video::ETC_REPEAT;
-	}
-	OverrideMaterial2D=InitMaterial2D;
+	InitMaterial2D.AntiAliasing = video::EAAM_OFF;
+	InitMaterial2D.Lighting = false;
+	InitMaterial2D.ZWriteEnable = video::EZW_OFF;
+	InitMaterial2D.ZBuffer = video::ECFN_DISABLED;
+	InitMaterial2D.UseMipMaps = false;
+	InitMaterial2D.forEachTexture([] (video::SMaterialLayer &tex) {
```
sfan5, line 105:
```suggestion
	InitMaterial2D.forEachTexture([] (auto &tex) {
```

include/EMaterialProps.h
grorp: I renamed it because "flag" sounded like "boolean value" to me, but not all of them are boolean values. Wasn't really necessary though.

grorp:
> Before I merge this can you drop the "(IrrlichtMt side)" from the commit messages?

Done.

sfan5:
Merged https://github.com/minetest/irrlicht/compare/393793f...c84d8ac

PR #205, opened 2023-06-25
Some build cleanups

nerzhul:
first, drop all #ifndef #define and use #pragma once everywhere (it's already used in code we added)
then, reduce some IRR_COMPILE_WITH uses where it's not needed because cmake already wrap the thing properly

@sfan5 tell me if android has special things, i should not have removed much useful headers, but maybe android is not using cmake and i was a bit ambitious


nerzhul:
hmm interesting windows is able to read linux files; ok that sound harder than expected but not impossible :D

nerzhul:
question @sfan5 should be detect OS from cmake as it's source of truth ? i ask for this because this clean leads to perform more cleanups on the build part which will permit to solve easily some red we have here

source/Irrlicht/CMakeLists.txt
```diff
@@ -457,6 +456,10 @@ add_library(IRROTHEROBJ OBJECT
 	os.cpp
 )
 
+if(LINUX_PLATFORM)
```
numberZero, line 458: Maybe check the `DEVICE` variable instead?

nerzhul: i can, but as we just setuped it for purposes like this i thought we can :)

nerzhul: if you look code, ti's enabled with EMSCRIPTEN or in default case for the device (L98 and L105)

numberZero: > if you look code

if you look at `git blame`...

> i can, but as we just setuped it for purposes like this i thought we can :)

don’t understand what you mean, sorry.

What I mean is, this `if(LINUX_PLATFORM)` duplicates a part of the device selection logic (which is at lines 83–117 currently).

numberZero: Also, `CIrrDeviceLinux` is a misnomer. It’s an X11 device. It won’t work on Emscripten. (not that Emscripten is officially supported but...)

nerzhul: okay i'll study that part

nerzhul:
replaced with #225

PR #209, opened 2023-07-02
Use SDL2 by default

### LINKED ISSUE #206: When default to SDL?
Basically, there are 2 options:
1. When the old device is dropped for a platform. Or,
2. Before that.

In either case, it should be done well before the next release. And in the case 2, it is also possible to revert the default before the release, e.g. if some unexpectedly hard issues are uncovered. Other than that, I see no point in waiting. Bug reports won’t appear until people will actually *use* it, on many very different systems.

numberZero:
Closes #206.

### DIFF
--- a/scripts/ci-build-mingw.sh
+++ b/scripts/ci-build-mingw.sh
@@ -46,9 +46,7 @@ if [ $with_sdl -eq 1 ]; then
 		-DCMAKE_PREFIX_PATH=$libs/sdl2/lib/cmake
 	)
 else
-	tmp+=(
-		-DUSE_SDL2=OFF
-	)
+	tmp+=(-DUSE_SDL2=OFF)
 fi
 #[ $with_gl3 -eq 1 ] && tmp+=(-DENABLE_OPENGL=OFF -DENABLE_OPENGL3=ON)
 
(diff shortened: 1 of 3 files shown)


numberZero:
Mingw+SDL will apparently require uploading an SDL build to http://minetest.kitsunemimi.pw/. @sfan5 is that yours?

ndren (commented):
I think lowercase `sdl2` was the correct formatting in Brew (the way it was before): https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fsdl2

numberZero:
Rebased.
Now with all the CI already in place, it’s a very small change.

rubenwardy (approved):
I deleted my cmake cache and clean built, uses SDL as expected. EndeavourOS Linux

scripts/ci-build-mingw.sh
```diff
@@ -40,10 +40,16 @@ tmp=(
...
-)
+if [ $with_sdl -eq 1 ]; then
+	tmp+=(
+		-DUSE_SDL2=ON
+		-DCMAKE_PREFIX_PATH=$libs/sdl2/lib/cmake
+	)
+else
+	tmp+=(
+		-DUSE_SDL2=OFF
+	)
```
sfan5, line 51:
```suggestion
	tmp+=(-DUSE_SDL2=OFF)
```

PR #210, opened 2023-07-02
Include the unified driver by default

numberZero:
As Mesebox defaults to `opengl` and not `opengl3` it should be safe.


sfan5 (approved):
lgtm

sfan5:
Any reason this is still a draft?

numberZero:
> Any reason this is still a draft?

Don’t remember what was the reason. Don’t see any now.

PR #224, opened 2023-07-20
Switch to pragma & add some things on .gitignore

nerzhul:


.gitignore
```diff
@@ -19,3 +19,7 @@ scripts/glext.h
 *.dir/
 *.sln
 *visualstudio/
+
+# vscode cmake plugin
+build/.cmake
+build/compile_commands.json
```
sfan5, line 25: just blacklist the entire folder?

nerzhul: fixed, right color was buggy in vscode, i thought some folders were it git :)

PR #225, opened 2023-07-26
cleanup: only build CIrrDeviceWin32 on WIN32

nerzhul:
remove _IRR_COMPILE_WITH_WINDOWS_DEVICE_ variable. It's now only compiled on win32


source/Irrlicht/CMakeLists.txt
```diff
@@ -139,6 +142,10 @@ if(NOT (BSD OR SOLARIS OR EMSCRIPTEN))
 	add_definitions(-D_IRR_COMPILE_WITH_JOYSTICK_EVENTS_)
 endif()
 
+# EGL
+
+option(ENABLE_EGL "Enable EGL" FALSE)
```
sfan5, line 147: this shouldn't be an option, it's not something the user can choose

source/Irrlicht/CMakeLists.txt
```diff
@@ -139,6 +142,10 @@ if(NOT (BSD OR SOLARIS OR EMSCRIPTEN))
 	add_definitions(-D_IRR_COMPILE_WITH_JOYSTICK_EVENTS_)
 endif()
 
+# EGL
+
+set(ENABLE_EGL OFF)
```
sfan5, line 147: there are `set(ENABLE_EGL)` calls above this line, the value will be reset

nerzhul: ah yeah right it's too upper my bad

nerzhul: fixed

nerzhul:
don't hesitate to squash and edit commit msg

PR #226, opened 2023-08-15
CIrrDeviceLinux: Fix type for NET_WM_PID XChangeProperty

jrtc27:
This uses format=32 which in X11's API means a C type of long (with restricted range when > 32 bits). pid_t is of unknown type, though on Linux and FreeBSD it's a 32-bit type, so does not have the same size as long on 64-bit architectures, and thus XChangeProperty reads outside its bounds. Fix this by casting to and passing a long.


Desour (approved):
Thanks!

PR #228, opened 2023-08-26
Deduplicate KP_5 from KP_HOME

kawogi:
I had issues with setting the keybindings in minetest.

Both, the keypad 7 key (Home) and the keypad 5 key were mapped to `Home` (irrespective of the Num-lock status).

I remapped `…BEGIN` (which seems to be Num5) to `KEY_NUMPAD5`.

After that change I was able to properly differentiate the keybindings in minetest.

(Tested on Linux Mint 21.1, Gnome 42.9, Wayland)

### DIFF
--- a/source/Irrlicht/CIrrDeviceLinux.cpp
+++ b/source/Irrlicht/CIrrDeviceLinux.cpp
@@ -1485,7 +1485,7 @@ void CIrrDeviceLinux::createKeyMap()
 	KeyMap.push_back(SKeyMap(XK_Next, KEY_NEXT));
 	KeyMap.push_back(SKeyMap(XK_Page_Down, KEY_NEXT));
 	KeyMap.push_back(SKeyMap(XK_End, KEY_END));
-	KeyMap.push_back(SKeyMap(XK_Begin, KEY_HOME));
+	KeyMap.push_back(SKeyMap(XK_Begin, KEY_NUMPAD5));
 	KeyMap.push_back(SKeyMap(XK_Num_Lock, KEY_NUMLOCK));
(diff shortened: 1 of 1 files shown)


source/Irrlicht/CIrrDeviceLinux.cpp
```diff
@@ -1485,7 +1485,7 @@ void CIrrDeviceLinux::createKeyMap()
 	KeyMap.push_back(SKeyMap(XK_Next, KEY_NEXT));
 	KeyMap.push_back(SKeyMap(XK_Page_Down, KEY_NEXT));
 	KeyMap.push_back(SKeyMap(XK_End, KEY_END));
-	KeyMap.push_back(SKeyMap(XK_Begin, KEY_HOME));
+	KeyMap.push_back(SKeyMap(XK_Begin, KEY_NUMPAD5));
```
sfan5, line 1488: shouldn't this still be KEY_HOME? it's not the one on the numpad

kawogi: Good question, I just did it for symmetry without further thinking about it. How was the the old binding to `KEY_HOME` chosen?

It looks like `XK_Begin` only exists for symmetry, but I failed to find a use for it anywhere.

https://sourceforge.net/p/nebuladevice/mailman/message/5216679/
https://stackoverflow.com/questions/25259852/what-is-the-begin-key
https://bugs.java.com/bugdatabase/view_bug?bug_id=4850137 (the most helpful so far)

Maybe it is a solution to just remove the binding? It looks like there's no keyboard that actually has this key.

sfan5: Hmm, looks like you're right. This can stay as-is.

PR #230, opened 2023-08-31
Use localtime_r over localtime

sfan5:


sfan5:
I think this code can actually just be removed, will check later.

Issue #168, opened 2023-02-28
BindingGenerator conflates core and extension constants


numberZero:
BindingGenerator conflates constant names with different vendor suffixes. But these constants aren’t always equal. I’ve written a tool and it found 18 discrepancies in `glcorearb.h` and `gl2ext.h` together.

Khronos/Khronos discrepancy:

* GL_UNSIGNED_INT_10_10_10_2: 0x8036 (core), 0x8df6 (OES)
* GL_HALF_FLOAT: 0x140b (core), 0x8d61 (OES)
* GL_GEOMETRY_VERTICES_OUT: 0x8916 (core), 0x8dda (ARB)
* GL_GEOMETRY_INPUT_TYPE: 0x8917 (core), 0x8ddb (ARB)
* GL_GEOMETRY_OUTPUT_TYPE: 0x8918 (core), 0x8ddc (ARB)

Khronos/major vendor discrepancy:

* GL_DEPTH_COMPONENT32F: 0x8cac (core), 0x8dab (NV)
* GL_DEPTH32F_STENCIL8: 0x8cad (core), 0x8dac (NV)
* GL_UNSIGNED_INT64: 0x140f (ARB, NV), 0x8bc2 (AMD)

Other:

* GL_TEXTURE_WIDTH: 0x1000 (core), 0x8bd2 (QCOM)
* GL_TEXTURE_HEIGHT: 0x1001 (core), 0x8bd3 (QCOM)
* GL_TEXTURE_INTERNAL_FORMAT: 0x1003 (core), 0x8bd5 (QCOM)
* GL_TEXTURE_DEPTH: 0x8071 (core), 0x8bd4 (QCOM)
* GL_RENDERBUFFER_SAMPLES: 0x8cab (core, ANGLE, APPLE, EXT, NV), 0x9133 (IMG)
* GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE: 0x8d56 (core, ANGLE, APPLE, EXT, NV), 0x9134 (IMG)
* GL_MAX_SAMPLES: 0x8d57 (core, ANGLE, APPLE, EXT, NV), 0x9135 (IMG)
* GL_TEXTURE_SAMPLES: 0x9106 (core), 0x9136 (IMG)
* GL_TEXTURE_TARGET: 0x1006 (core), 0x8bda (QCOM)
* GL_CONSERVATIVE_RASTERIZATION: 0x83fe (INTEL), 0x9346 (NV)

[...]

PR #232, opened 2023-09-11
Remove direct libGL linkage from OpenGL 3 and OGLES 2 Drivers

calebabutler:
This pull request does three things: 1. It changes the OpenGL 3 video driver to use a Core 3.3 Profile; 2. It decouples the legacy OpenGL driver from the modern OpenGL 3 and OGLES 2 drivers; that is, it allows you to compile Irrlicht without it; and 3. It removes the need to directly link to libGL when only compiling the modern OpenGL 3 driver or OGLES 2 driver without the legacy OpenGL driver.

The current OpenGL 3 video driver does not work with a Core profile, since the Core profile requires a Vertex Array Object and a Vertex Buffer Object to be bound to draw anything, which is not always the case currently. So the (inefficient) way that I made it work was I created a single global Vertex Array Object and a single global Vertex Buffer Object that are used when the renderer does not have other ones to use. The way to make this driver more efficient would be to reuse Vertex Array Objects and Vertex Buffer Objects as much as possible, hopefully reducing the communication between the CPU and GPU as much as possible.


source/Irrlicht/OpenGL/Driver.h
```diff
@@ -404,11 +413,11 @@ namespace video
...
 
 		unsigned QuadIndexCount;
-		GLuint QuadIndexBuffer = 0;
+		unsigned int QuadIndexBuffer = 0;
 		void initQuadsIndices(int max_vertex_count = 65536);
 
-		void debugCb(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message);
-		static void APIENTRY debugCb(GLenum source, GLenum type, GLuint id, GLenum severity, GLsizei length, const GLchar *message, const void *userParam);
+		void debugCb(unsigned int source, unsigned int type, unsigned int id, unsigned int severity, int length, const char *message);
+		static void APIENTRY debugCb(unsigned int source, unsigned int type, unsigned int id, unsigned int severity, int length, const char *message, const void *userParam);
```
sfan5, line 420: changing all these types certainly doesn't seem ideal.

do our gl bindings not provide them? if so that should be fixed instead
see also #168, maybe we should switch to something else?

sfan5:
cc @numberZero

calebabutler:
I would be happy using glad. It would make things pretty simple and we would keep a lot more bindings. Right now the gl bindings don't provide those types.

source/Irrlicht/OpenGL/Driver.h
numberZero: > changing all these types certainly doesn't seem ideal.

And isn’t even required to avoid *linking* to `libGL`. They only need headers. (likewise with constants btw but they are fine when used via `GL.`)

sfan5:
Superseded.

PR #233, opened 2023-09-13
[NOSQUASH] Remove the unnecessary sort in CNullDriver::addTexture

Desour:
TL;DR: A performance fix for faster client startup.

I've done some little profiling on client connect. Here's a flame graph showing Game::startup of a session where it took about 40 seconds to connect to a server:

graph_long_before

The `std::__introsort_loop` is the sort of CNullDriver::Textures that we do for each texture. This sorting is unnecessary.

CNullDriver::Textures holds an irr::core::array that is possibly sorted by texture name. Textures are accessed via binary search (see irrArray.h), which always sorts anyway. The comment that was already there also explains this.
CNullDriver::Textures is also not used by any child class, FYI.

Benchmarks (of Client::afterContentReceived, and Game::startup in parenthesis):
(I haven't run these often. Values are rounded.)

lightweight scenario (singleplayer mtg + mesecons):
before: 1.1 s (2.0 s)
after: 0.9 (1.8)

heavy scenario (some server):
before: 35.9 s (44.4 s)
after: 5.0 s (12.7 s)

### DIFF
--- a/include/IVideoDriver.h
+++ b/include/IVideoDriver.h
@@ -265,23 +265,9 @@ namespace video
 		IReferenceCounted::drop() for more information. */
 		virtual ITexture* getTexture(io::IReadFile* file) =0;
 
-		//! Returns a texture by index
-		/** \param index: Index of the texture, must be smaller than
-		getTextureCount() Please note that this index might change when
-		adding or removing textures
-		\return Pointer to the texture, or 0 if the texture was not
-		set or index is out of bounds. This pointer should not be
-		dropped. See IReferenceCounted::drop() for more information. */
-		virtual ITexture* getTextureByIndex(u32 index) =0;
-
 		//! Returns amount of textures currently loaded
 		/** \return Amount of textures currently loaded */
 		virtual u32 getTextureCount() const = 0;
-
-		//! Renames a texture
-		/** \param texture Pointer to the texture to rename.
-		\param newName New name for the texture. This should be a unique name. */
-		virtual void renameTexture(ITexture* texture, const io::path& newName) = 0;
 
 		//! Creates an empty texture of specified size.
 		/** \param size: Size of the texture.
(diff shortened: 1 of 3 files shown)


sfan5:
I don't think we use this but this sort call can be removed too:
grafik

Desour:
`renameTexture` is actually broken, if I'm not wrong, because `Textures` is not flagged as unsorted before the `Textures.sort()` call.
I've removed the function now, together with another unused not useful one.

appgurueu:
(For the future) could this be refactored to use an unordered map?

sfan5:
As I just found out [here] texture names are apparently not an one-to-one mapping to textures :thinking: 
So, maybe.


[here]: https://irc.minetest.net/minetest-dev/2023-09-13#i_6113112

Desour:
If they are not supposed to be unique, we could still use an unordered multimap.
It would make sense imo, I just didn't do it because I wanted the change to be simple. (Also, if I change the header, I have to recompile half of minetest.)

SmallJoker:
The Irrlicht situation, visualized:
xkcd Standards

 @BeverlyCode Please feel free to forward such patches to upstream Irrlicht so that all forks *could* benefit from it. Fixes in private forks do not help anyone else.

appgurueu:
> @BeverlyCode Please feel free to forward such patches to upstream Irrlicht so that all forks _could_ benefit from it. Fixes in private forks do not help anyone else.

(Or just publicize your currently "private" fork so we can take a look).

Also feel free to responsibly disclose any vulnerabilities to us and/or Irrlicht as applicable.

appgurueu:
I looked through Irrlicht's commit history a bit, and it seemed to me like any recent security improvements (in loaders) stemmed from IrrlichtMT. Could you point to a specific commit, or as said, disclose the vulnerability to us? Are you sure there is presently a vulnerability in IrrlichtMT?

sfan5:
Our primary concern are security-relevant issues in the engine or IrrlichtMt.
You can disclose them as written in our [security policy](https://github.com/minetest/minetest/security/policy) and they will absolutely get looked at.

> and to be honest I don't really like any of you. 🤷

That is regrettable but it was your choice to comment here.

PR #237, opened 2023-09-18
File format loader fixes

sfan5:
I gave up on the .x loader since there's too many bugs (there are remaining ones).

Although I tried to be careful testing is welcome to check that I didn't accidentally break any valid models.

### DIFF
--- a/source/Irrlicht/CImageLoaderBMP.cpp
+++ b/source/Irrlicht/CImageLoaderBMP.cpp
@@ -45,20 +45,32 @@ bool CImageLoaderBMP::isALoadableFileFormat(io::IReadFile* file) const
 	return headerID == 0x4d42;
 }
 
+// UB-safe overflow check
+static inline bool overflowCheck(const void *base, size_t offset, const void *end)
+{
+	auto baseI = reinterpret_cast<uintptr_t>(base),
+		endI = reinterpret_cast<uintptr_t>(end);
+	return baseI > endI || offset >= (endI - baseI);
+}
+// check whether &p[0] to &p[_off - 1] can be accessed
+#define CHECKP(_off) if ((_off) < 0 || overflowCheck(p, _off, pEnd)) goto exit
+// same for d
+#define CHECKD(_off) if ((_off) < 0 || overflowCheck(d, _off, destEnd)) goto exit
 
 void CImageLoaderBMP::decompress8BitRLE(u8*& bmpData, s32 size, s32 width, s32 height, s32 pitch) const
 {
 	u8* p = bmpData;
+	const u8* pEnd = bmpData + size;
 	u8* newBmp = new u8[(width+pitch)*height];
 	u8* d = newBmp;
-	u8* destEnd = newBmp + (width+pitch)*height;
+	const u8* destEnd = newBmp + (width+pitch)*height;
 	s32 line = 0;
 
-	while (bmpData - p < size && d < destEnd)
+	while (p < pEnd && d < destEnd)
 	{
 		if (*p == 0)
 		{
-			++p;
+			++p; CHECKP(1);
 
 			switch(*p)
 			{
@@ -68,29 +80,28 @@ void CImageLoaderBMP::decompress8BitRLE(u8*& bmpData, s32 size, s32 width, s32 h
 				d = newBmp + (line*(width+pitch));
 				break;
 			case 1: // end of bmp
-				delete [] bmpData;
-				bmpData = newBmp;
-				return;
+				goto exit;
 			case 2:
-				++p; d +=(u8)*p;  // delta
-				++p; d += ((u8)*p)*(width+pitch);
-				++p;
+				++p; CHECKP(2);
+				d += (u8)*p; ++p; // delta
+				d += ((u8)*p)*(width+pitch); ++p;
 				break;
 			default:
 				{
 					// absolute mode
 					s32 count = (u8)*p; ++p;
 					s32 readAdditional = ((2-(count%2))%2);
-					s32 i;
 
-					for (i=0; i<count; ++i)
+					CHECKP(count); CHECKD(count);
+					for (s32 i=0; i<count; ++i)
 					{
 						*d = *p;
 						++p;
 						++d;
 					}
 
-					for (i=0; i<readAdditional; ++i)
+					CHECKP(readAdditional);
+					for (s32 i=0; i<readAdditional; ++i)
 						++p;
 				}
 			}
@@ -98,7 +109,9 @@ void CImageLoaderBMP::decompress8BitRLE(u8*& bmpData, s32 size, s32 width, s32 h
 		else
 		{
 			s32 count = (u8)*p; ++p;
+			CHECKP(1);
 			u8 color = *p; ++p;
+			CHECKD(count);
 			for (s32 i=0; i<count; ++i)
 			{
 				*d = color;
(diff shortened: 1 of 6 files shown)


source/Irrlicht/CImageLoaderBMP.cpp
```diff
@@ -45,20 +45,25 @@ bool CImageLoaderBMP::isALoadableFileFormat(io::IReadFile* file) const
 	return headerID == 0x4d42;
 }
 
+// check whether &p[0] to &p[_off - 1] can be accessed
+#define CHECKP(_off) if ((_off) < 0 || p + (_off) > pEnd) goto exit
+// same for d
+#define CHECKD(_off) if ((_off) < 0 || d + (_off) > destEnd) goto exit
```
numberZero, line 51: Technically computing `d + _off` is UB if it is actually out of bounds. And I’ve heard of compilers that elide such a check on this basis.

source/Irrlicht/CImageLoaderBMP.cpp
```diff
@@ -107,26 +113,29 @@ void CImageLoaderBMP::decompress8BitRLE(u8*& bmpData, s32 size, s32 width, s32 h
 		}
 	}
 
+exit:
 	delete [] bmpData;
 	bmpData = newBmp;
 }
 
+#define SHIFTED_COUNT(_count, _shift) ((_count) / 2 + (((_shift) == 0 || (_count) % 2 == 1) ? 1 : 0))
```
numberZero, line 121: What does it do?
And, why a macro? Wouldn’t inline static function be equally suitable and more readable?

source/Irrlicht/CImageLoaderBMP.cpp
```diff
@@ -155,15 +162,16 @@ void CImageLoaderBMP::decompress4BitRLE(u8*& bmpData, s32 size, s32 width, s32 h
 					s32 count = (u8)*p; ++p;
 					s32 readAdditional = ((2-((count)%2))%2);
 					s32 readShift = 4;
-					s32 i;
 
-					for (i=0; i<count; ++i)
+					CHECKP(SHIFTED_COUNT(count, readShift));
+					CHECKD(SHIFTED_COUNT(count, shift));
```
numberZero, line 167: What does this check?

source/Irrlicht/CImageLoaderBMP.cpp
```diff
@@ -262,12 +276,16 @@ IImage* CImageLoaderBMP::loadImage(io::IReadFile* file) const
...
 	s32 paletteSize = (header.BitmapDataOffset - pos) / 4;
+	paletteSize = core::clamp(paletteSize, 0, paletteAllocSize);
 
 	s32* paletteData = 0;
 	if (paletteSize)
 	{
-		paletteData = new s32[paletteSize];
+		// always allocate an 8-bit palette to ensure enough space
+		paletteData = new s32[paletteAllocSize];
+		memset(paletteData, 0, paletteAllocSize * sizeof(s32));
```
numberZero, line 288: IIRC `new s32[paletteAllocSize]()` or `new s32[paletteAllocSize]{}` would do the zero-initialization as well (unlike bare `new s32[paletteAllocSize]`)

source/Irrlicht/CXMeshFileLoader.cpp
```diff
@@ -17,14 +17,16 @@
 #endif
 //#define BETTER_MESHBUFFER_SPLITTING_FOR_X
 
+#define SET_ERR_AND_RETURN() do { ErrorState = true; return false; } while (0)
+
```
numberZero, line 21: I guess this is what exceptions are for.

numberZero (commented):
Nice to see security enhancements)

> Although I tried to be careful testing is welcome to check that I didn't accidentally break any valid models.

Isn’t that what unit tests are for?

source/Irrlicht/CImageLoaderBMP.cpp
sfan5: I tried to google this but didn't find anything. Any references?

source/Irrlicht/CImageLoaderBMP.cpp
sfan5: the code operates on 4 bits, this computes how many bytes it will touch given the current state

source/Irrlicht/CImageLoaderBMP.cpp
sfan5: check that `p`/`d` are big enough for which bytes the following loop touches

source/Irrlicht/CXMeshFileLoader.cpp
sfan5: Indeed, however I won't be adding them to this code.

sfan5:
> Isn’t that what unit tests are for?

True.
I'm not very motivated to add them to this code because we hopefully have to never touch it again when GLTF takes over all modelling needs inside MT.

source/Irrlicht/CImageLoaderBMP.cpp
Desour: >Wouldn’t inline static function be equally suitable and more readable?

Or a lambda in the function.

source/Irrlicht/SB3DStructs.h
```diff
@@ -25,6 +25,8 @@ struct SB3dChunk
 	SB3dChunk(const SB3dChunkHeader& header, long sp)
 		: length(header.size+8), startposition(sp)
 	{
+		if (length < 0)
+			length = 8;
```
Desour, line 29: Shouldn't this be `length = max(length, 8);`?

source/Irrlicht/CXMeshFileLoader.cpp
```diff
@@ -437,6 +438,8 @@ bool CXMeshFileLoader::readFileIntoMemory(io::IReadFile* file)
 		return false;
 	}
 	BinaryNumCount=0;
+	if (BinaryFormat)
+		End--; // remove null-terminator again, only needed for string-format
```
Desour, line 444: `End` looks exclusive to me. So `End = Buffer + size` seems reasonable for binary format.

Desour (commented):
It would be good to value-init headers before reading them, e.g. in the BMP loader, for the case read() doesn't read a full header.

We should at some point replace some of these loaders with libraries, e.g. `sdl2_image` for images.

Haven't tested. Unittests would be good.

source/Irrlicht/CImageLoaderBMP.cpp
Desour: https://en.cppreference.com/w/cpp/language/pointer
>Indirection through an invalid pointer value and passing an invalid pointer value to a deallocation function have undefined behavior. Any other use of an invalid pointer value has implementation-defined behavior. Some implementations might define that copying an invalid pointer value causes a system-generated runtime fault.

source/Irrlicht/CXMeshFileLoader.cpp
sfan5: is a mistake, will fix

source/Irrlicht/CImageLoaderBMP.cpp
numberZero: https://en.cppreference.com/w/c/language/operator_arithmetic
> The behavior is defined only if both the original pointer and the result pointer are pointing at elements of the same array or one past the end of that array.

Or to be pedantic, [expr.add].4:
> When an expression that has integral type is added to or subtracted from a pointer, the result has the type
of the pointer operand. If the expression $P$ points to element $x[i]$ of an array object $x$ with $n$ elements,
the expressions $P + J$ and $J + P$ (where $J$ has the value $j$) point to the (possibly-hypothetical) element
$x[i + j]$ if $0 ≤ i + j ≤ n$; **otherwise, the behavior is undefined.** Likewise, the expression $P - J$ points to the
(possibly-hypothetical) element $x[i − j]$ if $0 ≤ i − j ≤ n$; otherwise, the behavior is undefined.

(emphasis mine)

sfan5: Fixed using `uintptr_t` casts

PR #240, opened 2023-09-20
Delete obsolete stuff

numberZero:
Delete `bin`, `lib` and `doc`, keeping only the license from the latter.


sfan5:
The build system places its files there so at least that has to be adjusted too.

Desour:
>The build system places its files there so at least that has to (or should) be adjusted too.

Tested (with out-of-tree and in-tree build). The paths are created automatically. I don't see an issue.

`irr-readme.txt` has some author names. Should we keep them?

numberZero:
> The build system places its files there so at least that has to (or should) be adjusted too.

At least on Linux, it places the files in identically named directories in the build directory. You don’t build in-tree, do you?

> `irr-readme.txt` has some author names. Should we keep them? Or link to https://irrlicht.sourceforge.io/?page_id=301 somewhere maybe?

Maybe. But, the comments are largely historical. And, note this: “this list hasn't been updated in a while, but they are often mentioned in changes.txt”. And there is no changes.txt.

I think a simple link to the original project may be more appropriate. Something like “Based on the [Irrlicht Engine](https://irrlicht.sourceforge.io/)”.

PR #241, opened 2023-09-22
ComboBox API additions for minetest#13814

grorp:
This PR makes improvements to the combobox API for minetest/minetest#13814.

- Allow reliably detecting when the user opens a combobox.
- Allow preventing comboboxes from opening.
- Allow sending a combobox change event.

### DIFF
--- a/include/IGUIComboBox.h
+++ b/include/IGUIComboBox.h
@@ -52,6 +52,10 @@ namespace gui
 		//! Sets the selected item. Set this to -1 if no item should be selected
 		virtual void setSelected(s32 idx) = 0;
 
+		//! Sets the selected item and emits a change event.
+		/** Set this to -1 if no item should be selected */
+		virtual void setAndSendSelected(s32 idx) = 0;
+
 		//! Sets text justification of the text area
 		/** \param horizontal: EGUIA_UPPERLEFT for left justified (default),
 		EGUIA_LOWERRIGHT for right justified, or EGUIA_CENTER for centered text.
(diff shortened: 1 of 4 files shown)


srifqi (approved):

source/Irrlicht/CGUIComboBox.cpp
```diff
@@ -451,8 +460,19 @@ void CGUIComboBox::openCloseMenu()
 	}
 	else
 	{
-		if (Parent)
+		if (Parent) {
+			SEvent event;
+			event.EventType = EET_GUI_EVENT;
+			event.GUIEvent.Caller = this;
+			event.GUIEvent.Element = 0;
```
numberZero, line 467: nullptr

include/IGUIComboBox.h
```diff
@@ -52,6 +52,10 @@ namespace gui
 		//! Sets the selected item. Set this to -1 if no item should be selected
 		virtual void setSelected(s32 idx) = 0;
 
+		//! Sets the selected item and emits a change event.
+		/** Set this to -1 if no item should be selected */
+		virtual void setAndSendSelected(s32 idx) = 0;
```
sfan5, line 57: why doesn't `setSelected` just do this?
I'm not saying you have to change the PR in that fashion, but it would be more logical.

grorp: I agree that it would be better to have `setSelected` do this, but `setSelected` is used a lot in `CGUIComboBox.cpp` and I'd prefer not to rewrite that file.

PR #244, opened 2023-09-30
Add tests for image loaders

numberZero:
This PR adds simple tests for BMP, PNG and TGA image loaders. Image dimensions and content are tested.

Image size is deliberately odd, to capture bugs related to unusual image sizes (apparently BMP loader has this exact problem with 16bpp images).

I commented out tests that fail currently to reduce noise. These should be enabled on fixing underlying problems.

I only added testing to the linux-gl build as the tests look platform-independent currently. Not sure whether that’s a good idea though.


sfan5 (approved):
after importing irrlicht into mt we should probably decide which testing framework these should use in the future.

numberZero:
Updated. “ImageLoaderTGA-24bpp_rle_up” fails due to 4a7d3de89a88db7e2f82d80630ac89b5f66a3ed9.

numberZero:
> “ImageLoaderTGA-24bpp_rle_up” fails due to [4a7d3de](https://github.com/minetest/irrlicht/commit/4a7d3de89a88db7e2f82d80630ac89b5f66a3ed9).

Fixed with c47300b23cd5a368081f1243dfad4882bdc7c6b7

PR #248, opened 2023-10-04
Reformat the code

numberZero:
This is a (hopefully final) prerequisite for minetest/minetest#13642.

I used `.clang-format` from Minetest but altered it a bit. In particular, `clang-format` isn’t all that good on wrapping long lines, so I removed line length limit to avoid ugly continuation.

Now, I don’t claim it is *all* pretty and nice now, but it’s much more consistent with both Minetest and itself.


numberZero:
Reapplied.

Desour (commented):
Does what it says (apart from those 2 spaces).

How hard would it be to replace the "Strip non-leading tabs" commit with one that replaces all the tabs with the right amount of spaces? It's a bit sad that we're loosing the alignment.

numberZero:
> It's a bit sad that we're loosing the alignment.

Alignment?
Oh wait, you’re likely still using a monospaced font. Okay...

> How hard would it be to replace the "Strip non-leading tabs" commit with one that replaces all the tabs with the right amount of spaces?

Doing exactly what you suggest is possible but won’t help as great deal of these non-leading tabs wasn’t even legitimate alignment. Up to [this sample](https://github.com/minetest/irrlicht/pull/248/commits/cbe19be4d706794960942658a777911e3cc553f5#diff-b2eb00ad51aa81194781d13352ed77bfa7ab293709101798368ed0dd48ea2b15) in include/SMaterialLayer.h. Manually working through that is possible (it’s just 1k lines) but out of scope of this PR.

Saying that, now I think [that commit](https://github.com/minetest/irrlicht/pull/248/commits/cbe19be4d706794960942658a777911e3cc553f5) wasn’t at all good, due to e.g. code samples in comments. I removed it from the branch as it would be much easier to reapply its good part than to undo its bad part later.

Desour (approved):
LGTM.

Desour:
>Oh wait, you’re likely still using a monospaced font. Okay...

Yes, I just find it's neat.

>Manually working through that is possible (it’s just 1k lines) but out of scope of this PR.

I thought more about writing a script that does it.

>Saying that, now I think [that commit](https://github.com/minetest/irrlicht/pull/248/commits/cbe19be4d706794960942658a777911e3cc553f5) wasn’t at all good, due to e.g. code samples in comments. I removed it from the branch as it would be much easier to reapply its good part than to undo its bad part later.

Good solution for now. :+1:

numberZero:
Reapplied.
Yes, it has to be reapplied when pretty much any other PR is merged. For the obvious reason. Maybe it’s better to split it in parts?

sfan5:
I think we should do this as the very last step before importing it into the MT repo since at that point the ability to easily backport upstream commits and all PRs on this repo are lost anyway.

numberZero:
What else holds the import, then?

sfan5:
I think we can also use this opportunity to move all files from `source/Irrlicht` to just `src`. thoughts?

Desour:
@numberZero I've seen you've deleted some of your fork repos. If you don't want to work on this anymore, we'll adopt this. :-) I hope you're doing fine though.

Desour:
#295

>I think we can also use this opportunity to move all files from source/Irrlicht to just src. thoughts?

Included in new PR.

PR #249, opened 2023-10-05
Fix character encoding conversion issues

### LINKED ISSUE #216: SDL 2 on Windows: Problems with Unicode/UTF-8

grorp:
Continuation of #221. Fixes #216 and fixes minetest/minetest#13646.

The Irrlicht string conversion functions are not broken, they are just the wrong kind of conversion functions: Irrlicht uses `wide <-> multibyte` conversion functions in places where it should actually use `wide <-> UTF-8` conversion functions. This usually works because the locale-defined multibyte encoding is UTF-8 in many environments, but sometimes, it doesn't.

This PR introduces new `wide <-> UTF-8` conversion functions, `utf8ToWString` and `wStringToUTF8`, and uses them where appropriate. `wStringToMultibyte` is not used anymore and has been removed, `multibyteToWString` has been kept because it is still used in one place.

EDIT: btw, "AutomatedTest" doesn't even compile on Windows.

## To do

This PR is Ready for Review. Before merging, 69622fe641017a10eaf05d0f36186fea424d80ec should be reverted.

## How to test

Compile Minetest on Windows with and without SDL. Set a weird combination of language settings. Verify that Unicode text is displayed/copied/whatever-ed correctly.

### DIFF
--- a/examples/AutomatedTest/test_string.cpp
+++ b/examples/AutomatedTest/test_string.cpp
@@ -168,22 +168,31 @@ static void test_methods()
 
 static void test_conv()
 {
-	// assumes Unicode and UTF-8 locale
-	setlocale(LC_CTYPE, "");
+	// should be locale-independent
 
 	stringw out;
-	multibyteToWString(out, "†††");
+	utf8ToWString(out, "†††");
 	UASSERTEQ(out.size(), 3);
 	for (int i = 0; i < 3; i++)
 		UASSERTEQ(static_cast<u16>(out[i]), 0x2020);
+
 	stringc out2;
-	wStringToMultibyte(out2, L"†††");
+	wStringToUTF8(out2, L"†††");
 	UASSERTEQ(out2.size(), 9);
 	for (int i = 0; i < 3; i++) {
 		UASSERTEQ(static_cast<u8>(out2[3*i]), 0xe2);
 		UASSERTEQ(static_cast<u8>(out2[3*i+1]), 0x80);
 		UASSERTEQ(static_cast<u8>(out2[3*i+2]), 0xa0);
 	}
+
+	// assumes Unicode and UTF-8 locale
+	setlocale(LC_CTYPE, "");
+
+	stringw out3;
+	multibyteToWString(out3, "†††");
+	UASSERTEQ(out3.size(), 3);
+	for (int i = 0; i < 3; i++)
+		UASSERTEQ(static_cast<u16>(out3[i]), 0x2020);
 }
 
 void test_irr_string()
--- a/examples/AutomatedTest/test_string.cpp
+++ b/examples/AutomatedTest/test_string.cpp
@@ -168,7 +168,7 @@ static void test_methods()
 
 static void test_conv()
 {
-	// should be locale-independent
+	// locale-independent
 
 	stringw out;
 	utf8ToWString(out, "†††");
@@ -185,8 +185,8 @@ static void test_conv()
 		UASSERTEQ(static_cast<u8>(out2[3*i+2]), 0xa0);
 	}
 
-	// assumes Unicode and UTF-8 locale
-	setlocale(LC_CTYPE, "");
+	// locale-dependent
+	setlocale(LC_CTYPE, "C.UTF-8");
 
 	stringw out3;
 	multibyteToWString(out3, "†††");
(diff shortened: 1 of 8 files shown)


examples/AutomatedTest/test_string.cpp
```diff
@@ -168,22 +168,31 @@ static void test_methods()
...
+	wStringToUTF8(out2, L"†††");
 	UASSERTEQ(out2.size(), 9);
 	for (int i = 0; i < 3; i++) {
 		UASSERTEQ(static_cast<u8>(out2[3*i]), 0xe2);
 		UASSERTEQ(static_cast<u8>(out2[3*i+1]), 0x80);
 		UASSERTEQ(static_cast<u8>(out2[3*i+2]), 0xa0);
 	}
+
+	// assumes Unicode and UTF-8 locale
+	setlocale(LC_CTYPE, "");
```
numberZero, line 189: Maybe this?
```suggestion
	setlocale(LC_CTYPE, "C.UTF-8");
```

include/irrString.h
```diff
@@ -975,50 +965,29 @@ static size_t multibyteToWString(string<wchar_t>& destination, const char* sourc
...
+	return destination.size();
 }
 
-//! Same as multibyteToWString, but the other way around
-static inline size_t wStringToMultibyte(string<c8>& destination, const wchar_t* source)
+inline size_t utf8ToWString(stringw &destination, const stringc &source)
 {
-	const u32 s = source ? (u32)wcslen(source) : 0;
-	return wStringToMultibyte(destination, source, s);
+	return utf8ToWString(destination, source.c_str());
```
numberZero, line 978: Can a `stringc` contain embedded nulls? If so, this shouldn’t delegate but use another `from_bytes`:
`conv.from_bytes(source.begin(), sorce.end())`

grorp: This seems to suggest it can't:

https://github.com/minetest/irrlicht/blob/5b2f1927d0ccc59173e00f3276b33e08b1ca748c/include/irrString.h#L789-L798

include/irrString.h
```diff
@@ -913,46 +923,26 @@ class string
...
+\return The number of wide characters written to destination, not including the eventual terminating null character or -1 when conversion failed
 
-//! Convert multibyte string to wide-character string
-/** Wrapper around mbstowcs from standard library, but directly writing to Irrlicht string class.
-What the function does exactly depends on the LC_CTYPE of the current c locale.
-\param destination Wide-character string receiving the converted source
-\param source multibyte string
-\return The number of wide characters written to destination, not including the eventual terminating null character  or -1 when conversion failed. */
-static inline size_t multibyteToWString(string<wchar_t>& destination, const char* source)
+This function should not be used and is only kept for "CGUIFileOpenDialog::pathToStringW". */
```
sfan5, line 933: can we maybe just drop this function?
or wait for https://github.com/minetest/minetest/pull/13512 so we can drop the class entirely...

source/Irrlicht/CMakeLists.txt
```diff
@@ -142,7 +144,7 @@ endif()
 # OpenGL
 
 if(USE_SDL2)
-	option(ENABLE_OPENGL3 "Enable OpenGL 3+" TRUE)
+	option(ENABLE_OPENGL3 "Enable OpenGL 3+" FALSE)
```
sfan5, line 147: rebase mistake?

grorp: I temporarily disabled the new driver to make CI pass: 69622fe641017a10eaf05d0f36186fea424d80ec
I'll revert that commit and rebase this PR now that #250 is merged.

include/irrString.h
grorp: > can we maybe just drop this function?

How? If it is used?

> or wait for https://github.com/minetest/minetest/pull/13512 so we can drop the class entirely...

I'd like to avoid that, it could take a while until minetest/minetest#13512 is merged.

sfan5: You're right, we can always get rid of it later.

I did actually want to pick up https://github.com/minetest/minetest/pull/13512 and fix it because it's a good an necessary change, but I'm totally short on time right now.

sfan5:
Merged.

grorp:
Thanks!

Issue #236, opened 2023-09-18
TGA image loader has buffer overflows


sfan5:
Result of fuzzing with [AFL++](https://github.com/AFLplusplus/AFLplusplus). Setup/scripts are in [this branch](https://github.com/minetest/irrlicht/tree/afl).

This resulted in about 20 crashes (probably including duplicates) that look e.g. like this:
<pre>
==329007==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000005fd0 at pc 0x5626aacbcc07 bp 0x7ffcd7fe0ea0 sp 0x7ffcd7fe0e98
READ of size 4 at 0x602000005fd0 thread T0
    #0 0x5626aacbcc06 in irr::video::CColorConverter::convert8BitTo16Bit(unsigned char const*, short*, int, int, int const*, int, bool) /var/tmp/irr/source/Irrlicht/CColorConverter.cpp:106:32
    #1 0x5626aacb78e0 in irr::video::CImageLoaderTGA::loadImage(irr::io::IReadFile*) const /var/tmp/irr/source/Irrlicht/CImageLoaderTGA.cpp:189:6
    #2 0x5626aac8e4c6 in irr::video::CNullDriver::createImageFromFile(irr::io::IReadFile*) /var/tmp/irr/source/Irrlicht/CNullDriver.cpp:1155:41
    #3 0x5626aac6f11e in irr::video::CNullDriver::loadTextureFromFile(irr::io::IReadFile*, irr::core::string<char> const&) /var/tmp/irr/source/Irrlicht/CNullDriver.cpp:535:18
    #4 0x5626aac6f11e in irr::video::CNullDriver::getTexture(irr::core::string<char> const&) /var/tmp/irr/source/Irrlicht/CNullDriver.cpp:477:13
    #5 0x5626aa9c2897 in main /var/tmp/irr/examples/LoadTexture/main.cpp:22:23
    #6 0x7f9078227ccf  (/usr/lib/libc.so.6+0x27ccf) (BuildId: 316d0d3666387f0e8fb98773f51aa1801027c5ab)
    #7 0x7f9078227d89 in __libc_start_main (/usr/lib/libc.so.6+0x27d89) (BuildId: 316d0d3666387f0e8fb98773f51aa1801027c5ab)
    #8 0x5626aa883d24 in _start (/var/tmp/irr/bin/Linux/LoadTexture+0xbed24) (BuildId: 9d55f295133956d3)

[...]

PR #252, opened 2023-10-15
Backport upstream commits

sfan5:
addresses #236

someone should test macOS i guess


numberZero:
> addresses #236

Addresses but doesn’t fix. They did fix the overflows you put their nose to but missed nearly identical overflow just a few lines above.

sfan5:
Sure, we can still apply our own fixes. And maybe someone should notify upstream that they missed something.

PR #255, opened 2023-11-24
Android: Make ALooper_pollAll call always non-blocking

srifqi:
This PR tries to fix minetest/minetest#10842. See there for further discussion.

**Implementation details**

Original:

```cpp
((Focused && !Paused) || !Initialized) ? 0 : -1
```

Since we want Minetest to still run when the text box shows up (`Focused` is `false`) and when Minetest is minimised (`Paused` is `true`), we remove both variables from the formula.

```cpp
!Initialized ? 0 : -1
```

Because of this:

https://github.com/minetest/irrlicht/blob/85081d6fe0c422cf47f74714ee25562715528aa2/source/Irrlicht/Android/CIrrDeviceAndroid.cpp#L100-L101

we can remove `Initialized` from the formula.

```cpp
0
```

### DIFF
--- a/source/Irrlicht/Android/CIrrDeviceAndroid.cpp
+++ b/source/Irrlicht/Android/CIrrDeviceAndroid.cpp
@@ -106,7 +106,7 @@ bool CIrrDeviceAndroid::run()
 	s32 Events = 0;
 	android_poll_source* Source = 0;
 
-	while ((id = ALooper_pollAll(((Focused && !Paused) || !Initialized) ? 0 : -1, 0, &Events, (void**)&Source)) >= 0)
+	while ((id = ALooper_pollAll(0, 0, &Events, (void**)&Source)) >= 0)
 	{
 		if(Source)
 			Source->process(Android, Source);


grorp (commented):
Pausing Minetest completely when the app is paused doesn't make sense in multiplayer. It's also problematic in the mainmenu if the user is currently downloading a package from CDB. So +1 for the concept.

What about line 67 in the same file?

https://github.com/minetest/irrlicht/blob/b15dabe80ae50ea72513b01657880ad9b0881908/source/Irrlicht/Android/CIrrDeviceAndroid.cpp#L67

---------

To avoid an unnecessarily big increase in battery usage from this PR, maybe we could add code to Minetest to stop rendering while the app is paused? Something like this:

```patch
diff --git a/include/IrrlichtDevice.h b/include/IrrlichtDevice.h
index 6dcdb49..98c21f7 100644
--- a/include/IrrlichtDevice.h
+++ b/include/IrrlichtDevice.h
@@ -173,6 +173,8 @@ namespace irr
 		/** \return True if window is maximized. */
 		virtual bool isWindowMaximized() const = 0;
 
+		virtual bool isAppPaused() const { return false; };
+
 		//! Checks if the Irrlicht window is running in fullscreen mode
 		/** \return True if window is fullscreen. */
 		virtual bool isFullscreen() const = 0;
diff --git a/source/Irrlicht/Android/CIrrDeviceAndroid.cpp b/source/Irrlicht/Android/CIrrDeviceAndroid.cpp
index b1df885..745ed4f 100644
--- a/source/Irrlicht/Android/CIrrDeviceAndroid.cpp
+++ b/source/Irrlicht/Android/CIrrDeviceAndroid.cpp
@@ -193,6 +193,10 @@ bool CIrrDeviceAndroid::isWindowMinimized() const
 	return !Focused;
 }
 
+bool CIrrDeviceAndroid::isAppPaused() const {
+	return Paused;
+}
+
 void CIrrDeviceAndroid::closeDevice()
 {
 	ANativeActivity_finish(Android->activity);
diff --git a/source/Irrlicht/Android/CIrrDeviceAndroid.h b/source/Irrlicht/Android/CIrrDeviceAndroid.h
index 656c456..f5a88a5 100644
--- a/source/Irrlicht/Android/CIrrDeviceAndroid.h
+++ b/source/Irrlicht/Android/CIrrDeviceAndroid.h
@@ -36,6 +36,8 @@ namespace irr
 
 		virtual bool isWindowMinimized() const;
 
+		virtual bool isAppPaused() const;
+
 		virtual void closeDevice();
 
 		virtual void setResizable(bool resize = false);
```

```patch
diff --git a/src/client/game.cpp b/src/client/game.cpp
index cc88915474..4c9a8ae0bc 100644
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -1251,7 +1251,8 @@ void Game::run()
 		updateCamera(dtime);
 		updateSound(dtime);
 		processPlayerInteraction(dtime, m_game_ui->m_flags.show_hud);
-		updateFrame(&graph, &stats, dtime, cam_view);
+		if (!device->isAppPaused())
+			updateFrame(&graph, &stats, dtime, cam_view);
 		updateProfilerGraphs(&graph);
 
 		// Update if minimap has been disabled by the server
```

srifqi:
> What about line 67 in the same file?

I am not sure about that line. I have not tested it. Since that is run in initialisation step, it is probably fine to not change it.

-------

> To avoid an unnecessarily big increase in battery usage from this PR, maybe we could add code to Minetest to stop rendering while the app is paused? Something like this:

That looks good. The frame/window would not be shown anyway, but we still need the engine to run for other background tasks. It is probably better to also use conditional inclusion (`#ifdef` directive) since it is only used for Android devices.

sfan5:
> To avoid an unnecessarily big increase in battery usage from this PR, maybe we could add code to Minetest to stop rendering while the app is paused?

take a look at `isWindowActive`, it already enables exactly this

grorp (approved):
LGTM

I think it would be good to merge this together with minetest/minetest#14058.

PR #257, opened 2023-11-29
[not wip] This PR has imploded

jordan4ibanez:
WIP. Does what it says on the tin


rollerozxa:
3d models are no more, we can finally go home again

rubenwardy:
Why?

jordan4ibanez:
> Why?

Too many things pointing towards each other, too many things stepping on each other's toes. We start with the most complex loader first, GLTF. Then we put the OBJ loader back into the engine with the required changes.
Doing it this way stops literally everything else from getting in the way

rollerozxa:
The Irrlicht model format loader system is modular, you shouldn't need to do this just to get GLTF working. I assumed you were going to move it into Minetest and register them as external loaders but I don't know what the purpose would be of juggling around the loaders inside of IrrlichtMt like this.

jordan4ibanez:
> The Irrlicht model format loader system is modular, you shouldn't need to do this just to get GLTF working. I assumed you were going to move it into Minetest and register them as external loaders but I don't know what the purpose would be of juggling around the loaders inside of IrrlichtMt like this.

I'm probably going to have to change animation code to work with GLTF, I'm going into the modular loading system and modifying it and I don't need to be modifying 2 loaders at the same time, one being deprecated, to bring in a new one

jordan4ibanez:
So we're back to where we left of, excellent
screenshot_20231129_063426

PR #267, opened 2023-12-21
[no squash] Make equals method symmetric

savilli:
Fix minetest/minetest#14132
See https://github.com/minetest/minetest/issues/14132#issuecomment-1865302265
This fix is better as it also removes some nonsense.

### DIFF
--- a/include/irrMath.h
+++ b/include/irrMath.h
@@ -9,6 +9,7 @@
 #include <float.h>
 #include <stdlib.h> // for abs() etc.
 #include <limits.h> // For INT_MAX / UINT_MAX
+#include <type_traits>
 
 namespace irr
 {
@@ -17,9 +18,6 @@ namespace core
 
 	//! Rounding error constant often used when comparing f32 values.
 
-	const s32 ROUNDING_ERROR_S32 = 0;
-
-	const s64 ROUNDING_ERROR_S64 = 0;
 	const f32 ROUNDING_ERROR_f32 = 0.000001f;
 	const f64 ROUNDING_ERROR_f64 = 0.00000001;
 
@@ -170,30 +168,6 @@ namespace core
 		return ROUNDING_ERROR_f64;
 	}
 
-	template <>
-	inline s32 roundingError()
-	{
-		return ROUNDING_ERROR_S32;
-	}
-
-	template <>
-	inline u32 roundingError()
-	{
-		return ROUNDING_ERROR_S32;
-	}
-
-	template <>
-	inline s64 roundingError()
-	{
-		return ROUNDING_ERROR_S64;
-	}
-
-	template <>
-	inline u64 roundingError()
-	{
-		return ROUNDING_ERROR_S64;
-	}
-
 	template <class T>
 	inline T relativeErrorFactor()
 	{
@@ -212,13 +186,19 @@ namespace core
 		return 8;
 	}
 
-	//! returns if a equals b, taking possible rounding errors into account
-	template <class T>
-	inline bool equals(const T a, const T b, const T tolerance = roundingError<T>())
+	//! returns if a equals b, for types without rounding errors
+	template <class T, std::enable_if_t<std::is_integral<T>::value, bool> = true>
+	inline bool equals(const T a, const T b)
 	{
-		return (a + tolerance >= b) && (a - tolerance <= b);
+		return a == b;
 	}
 
+	//! returns if a equals b, taking possible rounding errors into account
+	template <class T, std::enable_if_t<std::is_floating_point<T>::value, bool> = true>
+	inline bool equals(const T a, const T b, const T tolerance = roundingError<T>())
+	{
+		return abs(a - b) <= tolerance;
+	}
 
 	//! returns if a equals b, taking relative error in form of factor
 	//! this particular function does not involve any division.
--- a/include/vector2d.h
+++ b/include/vector2d.h
@@ -75,36 +75,37 @@ public:
 		return *(&X+index);
 	}
 
-	//! sort in order X, Y. Equality with rounding tolerance.
+	//! sort in order X, Y.
 	bool operator<=(const vector2d<T>&other) const
 	{
-		return (X<other.X || core::equals(X, other.X)) ||
-				(core::equals(X, other.X) && (Y<other.Y || core::equals(Y, other.Y)));
+		return !(*this > other);
 	}
 
-	//! sort in order X, Y. Equality with rounding tolerance.
+	//! sort in order X, Y.
 	bool operator>=(const vector2d<T>&other) const
 	{
-		return (X>other.X || core::equals(X, other.X)) ||
-				(core::equals(X, other.X) && (Y>other.Y || core::equals(Y, other.Y)));
+		return !(*this < other);
 	}
 
-	//! sort in order X, Y. Difference must be above rounding tolerance.
+	//! sort in order X, Y.
 	bool operator<(const vector2d<T>&other) const
 	{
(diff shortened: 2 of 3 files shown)


include/vector2d.h
```diff
@@ -111,11 +111,10 @@ class vector2d
 	//! Checks if this vector equals the other one.
 	/** Takes floating point rounding errors into account.
 	\param other Vector to compare with.
-	\param tolerance Epsilon value for both - comparing X and Y.
 	\return True if the two vector are (almost) equal, else false. */
-	bool equals(const vector2d<T>& other, const T tolerance = (T)ROUNDING_ERROR_f32 ) const
```
savilli, line 116: Not a good default. Fortunately, this parameter is not used.

sfan5:
The changes look good and I'm convinced they should work but this still fails my test code:
```diff
diff --git a/source/Irrlicht/COBJMeshFileLoader.cpp b/source/Irrlicht/COBJMeshFileLoader.cpp
--- a/source/Irrlicht/COBJMeshFileLoader.cpp
+++ b/source/Irrlicht/COBJMeshFileLoader.cpp
@@ -2,6 +2,8 @@
 // This file is part of the "Irrlicht Engine".
 // For conditions of distribution and use, see copyright notice in irrlicht.h
 
+#undef NDEBUG
+#include <cassert>
 #include "COBJMeshFileLoader.h"
 #include "IMeshManipulator.h"
 #include "IVideoDriver.h"
@@ -241,6 +243,28 @@ IAnimatedMesh* COBJMeshFileLoader::createMesh(io::IReadFile* file)
 					currMtl->RecalculateNormals=true;
 				}
 
+				const auto pv = [] (const video::S3DVertex &v) {
+					printf("P = { %a , %a , %a }\n", v.Pos.X, v.Pos.Y, v.Pos.Z);
+					printf("N = { %a , %a , %a }\n", v.Normal.X, v.Normal.Y, v.Normal.Z);
+					printf("C = %#08x\n", v.Color.color);
+					printf("T = { %a , %a }\n", v.TCoords.X, v.TCoords.Y);
+				};
+				for (auto &it : currMtl->VertMap) {
+					if (it.first == v) {
+						assert(!(it.first < v));
+						assert(!(v < it.first));
+					} if (it.first < v) {
+						assert(!(v < it.first));
+					} else {
+						if (!(v < it.first)) {
+							pv(v);
+							puts("==");
+							pv(it.first);
+							assert(false);
+						}
+					}
+				}
+
 				int vertLocation;
 				auto n = currMtl->VertMap.find(v);
 				if (n != currMtl->VertMap.end())
```

with
```
P = { 0x1p-1 , 0x1.800086p-3 , 0x1.ffffbcp-2 }
N = { 0x1p+0 , 0x0p+0 , 0x0p+0 }
C = 0xffcccccc
T = { 0x0p+0 , 0x1.4p-2 }
==
P = { 0x1p-1 , 0x1.800086p-3 , 0x1.ffffbcp-2 }
N = { 0x1p+0 , 0x0p+0 , 0x0p+0 }
C = 0xffcccccc
T = { 0x0p+0 , 0x1.4p-2 }
```

sfan5:
Thinking further, isn't the problem that `<` is strict while `==` is fuzzy?

e.g. with whole numbers, tolerance of 1:
`1 == 2` but at the same time `1 < 2`

However if you look at the [relation requirements](https://en.cppreference.com/w/cpp/named_req/Compare) we have a fundamental problem. It's impossible to make transitivity work with fuzzy comparisons.

e.g. `1==2` and `2==3` but `!(1 == 3)`
or `!(1 < 2)` and `!(2 < 3)` but `1 < 3`

Basically `operator==` must not do fuzzy comparisons and anything depending on it needs to be reworked, right?

numberZero:
> isn't the problem that `<` is strict while `==` is fuzzy?

On numbers, both are strict (that’s why `equals` even exists). On vectors, both are fuzzy.

> It's impossible to make transitivity work with fuzzy comparisons.

Good point.

> or `!(1 < 2)` and `!(2 < 3)` but `1 < 3`

That’s not transitivity. It does break transitivity of equality though (defined as `!(a<b) && !(b<a)`), and thus the requirements you referenced. Ironically, that breaks the only reasonable use of `<` on vectors (using them as `map` keys)...

savilli:
Well, yes, we need to make comparison operators transitive as it's required for std::map. However, it would require removing approximate comparisons. It would be a more dangerous change because it's hard to say which code may rely on them.
I think it's better to change it in a separate PR and test that PR carefully. This PR definitely doesn't make comparison operators worse.

savilli:
Although, I took a quick look at where comparison operators are used in the code, and looks like it needs strict comparisons everywhere and not approximate ones.
I changed comparison operators and everything seems to be working. Your test code passes too. I guess it should be safe enough to commit.
Btw, your test code has a bug. `} if (it.first < v) {` should be replaced with `} else if (it.first < v) {`.

sfan5:
> Btw, your test code has a bug. `} if (it.first < v) {` should be replaced with `} else if (it.first < v) {`.

oops lol, was that the reason it failed all the time?

savilli:
> oops lol, was that the reason it failed all the time?

No, it also failed for valid reasons like `it.first == v` and `it.first < v` at the same time.

sfan5 (approved):
Looks good to me but I thought `COBJMeshFileLoader` depended on this to merge vertices properly?

appgurueu:
I don't see why it would, for sane models at least. Wouldn't any sane write the exact same coordinates for the exact same vertex?

savilli:
I couldn't find any .obj file where it could be a problem. There are very few files where different vertices are approximately equal in the first place. But even in those files, vertices don't need to be merged.
For example, in [home_workshop_machines_3dprinter_corexy.obj](https://github.com/mt-mods/home_workshop_modpack/blob/master/home_workshop_machines/models/home_workshop_machines_3dprinter_corexy.obj) there are `v -0.154688 0.281094 0.199376` and `v -0.154687 0.281094 0.199376` approximate equal vertices. The first is used in face `f 987/889/53 989/890/53 993/891/53 991/892/53` and the second is used in `f 1001/905/51 995/906/51 997/907/51 1005/908/51`. It makes more sense for them to be just square faces and not "square with merged corners" faces.

include/irrMath.h
```diff
@@ -212,14 +186,20 @@ namespace core
...
+		return a == b;
+	}
+
 	//! returns if a equals b, taking possible rounding errors into account
-	template <class T>
+	template <class T, std::enable_if_t<std::is_floating_point<T>::value, bool> = true>
 	inline bool equals(const T a, const T b, const T tolerance = roundingError<T>())
 	{
-		return (a + tolerance >= b) && (a - tolerance <= b);
+		return abs(a - b) <= tolerance;
```
paradust7, line 200: Is this supposed to be used for floating points?
The stdlib signature of abs is `int abs(int x)`, so this doesn't make sense.

Not sure why, but MSVC generates an ambiguity error here in my xr branch but not master (possibly because I've got C++20 enabled)

sfan5: it's defined for different float and int types:
* https://en.cppreference.com/w/cpp/numeric/math/fabs
* https://en.cppreference.com/w/cpp/numeric/math/abs

paradust7: Ah. It needs to be std::abs then to avoid picking up the integer version. And the includes need to be cstdio and cmath.

sfan5: Can you send a PR?

paradust7: sent: https://github.com/minetest/irrlicht/pull/293

PR #269, opened 2024-01-04
Add IrrlichtDevice::isWindowVisible for Android

grorp:
This PR adds a method called `isWindowVisible` to `IrrlichtDevice`. It returns false on Android if the app is paused and true in all other cases and on all other platforms.

This is necessary because `isWindowActive` and similar methods not only return false on Android when the app is paused, but also when it is merely "unfocused" (e.g. a text input dialog is open).

### DIFF
--- a/include/IrrlichtDevice.h
+++ b/include/IrrlichtDevice.h
@@ -177,6 +177,10 @@ namespace irr
 		/** \return True if window is fullscreen. */
 		virtual bool isFullscreen() const = 0;
 
+		//! Checks if the application is truly paused.
+		//! Currently, this can only happen on Android.
+		virtual bool isAppPaused() const { return false; };
+
 		//! Get the current color format of the window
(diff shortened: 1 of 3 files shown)


include/IrrlichtDevice.h
```diff
@@ -177,6 +177,10 @@ namespace irr
 		/** \return True if window is fullscreen. */
 		virtual bool isFullscreen() const = 0;
 
+		//! Checks if the application is truly paused.
+		//! Currently, this can only happen on Android.
+		virtual bool isAppPaused() const { return false; };
```
sfan5, line 182: thinking about other platforms: what if we make this `isWindowVisible`?

grorp: I hope that "app is paused == app is invisible" is also true in multi-window mode, but it seems to be as far as I've tested.

Will do.

PR #270, opened 2024-01-04
Update minimum CMake version

jordan4ibanez:
Cmake in engine currently very old.

Give it a few birthdays so we can use newer features.

We are missing the ability to add link libraries to object library targets. We'd like this for the glTF development.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.12)
 
 # >=3.9 enables IPO; >=3.11 prefers GLVND
 if(${CMAKE_VERSION} VERSION_LESS 3.11)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.12) 
 
 set(IRRLICHTMT_REVISION 13)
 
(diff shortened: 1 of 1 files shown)


appgurueu:
This depends on https://github.com/minetest/irrlicht/pull/272

sfan5:
What for? Are we missing any features?

JosiahWI:
We are missing the ability to add link libraries to object library targets. We'd like this for the glTF development, and I'd like it for the Irrlicht CMakeLists as well to help clean it up.

sfan5:
ok in that case:
* Debian 10: 3.13 :heavy_check_mark: 
* RHEL 8 derivates: seems to differ but 3.18 is available :heavy_check_mark: 
* Ubuntu 20.04: 3.16 :heavy_check_mark: 

verdict: we can do this

jordan4ibanez:
yay

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.12)
 
 # >=3.9 enables IPO; >=3.11 prefers GLVND
 if(${CMAKE_VERSION} VERSION_LESS 3.11)
```
sfan5, line 4: you can get rid of this check

jordan4ibanez: I did, the thing

jordan4ibanez:
Do we have to do anything else?

sfan5:
https://github.com/minetest/irrlicht/pull/270#issuecomment-1877944665:

> This depends on #272

PR #272, opened 2024-01-05
Update build.yml: Upgrade bionic -> jammy for non-SDL builds

appgurueu:
Non-SDL builds were (in constrast to SDL builds, which use jammy) still using bionic, which has reached end of standard support.

(BTW, I'm not sure what our policy is here. We're not using the oldest supported LTS version (that'd be 20.04 Focal Fossa). Is the intention to use the most recent LTS version?)


sfan5:
focal would be better since we have [workflows](https://github.com/minetest/minetest/blob/master/.github/workflows/linux.yml) in the MT repo that use the artifacts from here

appgurueu:
Some of the xvfb checks don't seem to complete:

Screenshot from 2024-01-05 20-13-32

sfan5:
Moving them back out of docker might fix it, who knows (revert https://github.com/minetest/irrlicht/commit/f3a1f9f6560c82974f2254d6518f9cd4f4e692b6).

.github/workflows/build.yml
```diff
@@ -167,7 +152,7 @@ jobs:
           CXX: ${{matrix.config.arch}}-w64-mingw32-g++
           extras: ${{matrix.config.extras}}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v2
```
sfan5, line 155: :thinking:

appgurueu:
By the way, I noticed that the test steps differ. On linux-gl it's just running the unit tests via `ctest`:

```yaml
      - name: Test
        run: |
          ctest --output-on-failure
```

whereas the other ones *don't* run the unit tests, but run

```yaml
      - name: Test (headless)
        run: |
          cd bin/Linux
          ./AutomatedTest null

      - name: Test (Xvfb)
        run: |
          cd bin/Linux
          LIBGL_ALWAYS_SOFTWARE=true xvfb-run ./AutomatedTest ogles2
```

MinGW and MSVC don't seem to test at all, MacOS doesn't run unit tests. Is this intended?

sfan5:
Yes/no/maybe, pick one. The different test methods we have are disjoint. :shrug:

appgurueu:
The SDL builds on 20.04 are failing because the SDL version on 20.04 is too old: https://wiki.libsdl.org/SDL2/SDL_MouseWheelEvent says `preciseX` / `preciseY` were added in 2.0.18, but Focal seems to have 2.0.10.

(Given that we want to support 20.04, isn't this an IrrlichtMT issue? Is there some way to get the same functionality as the precise values in 2.0.10, or should we just default to the less precise values if 2.0.18 isn't available?)

appgurueu:
Okay, after looking at this in a VM:

The error is `Could not initialize context: Could not create GL context: GLXBadFBConfig`. `glxinfo` claims a `Max compat profile version: 3.1`. We explicitly require 3.2. I think that might be the reason for the error; on my regular 22.04 Jammy I get `Max compat profile version: 4.6` from `glxinfo`.

I'm not sure how to fix this. Unless we get any users complaining, I would consider just upgrading CI to `jammy`.

sfan5:
We can move `linux-sdl-gl3` back to ubuntu 22, that's not the big issue.
But what about `linux-gles`?

appgurueu:
I basically get a blackscreen there, plus the "libEGL warning: DRI2: failed to authenticate". Googling that didn't bring up any fixes for Ubuntu that worked, though. The closest I could find was this [thread](https://irrlicht.sourceforge.io/forum/viewtopic.php?t=52826) on the Irrlicht forum by CuteAlien. However the suggested fix for Debian didn't work in my Ubuntu VM (though `es2gears` does work, albeit with the warning).

sfan5:
I checked using a Ubuntu 20.04 VM and the error doesn't appear but it instead gets stuck in setting the caption (?!):
```
#0  __lll_lock_wait (futex=futex@entry=0x55555558c690, private=128) at lowlevellock.c:52
#1  0x00007ffff6e710a3 in __GI___pthread_mutex_lock (mutex=0x55555558c690)
    at ../nptl/pthread_mutex_lock.c:80
#2  0x00007ffff7830a13 in _XOpenLC () from /lib/x86_64-linux-gnu/libX11.so.6
#3  0x00007ffff7830c0d in _XlcCurrentLC () from /lib/x86_64-linux-gnu/libX11.so.6
#4  0x00007ffff7830eaf in XwcTextListToTextProperty () from /lib/x86_64-linux-gnu/libX11.so.6
#5  0x00007ffff7f3d876 in irr::CIrrDeviceLinux::setWindowCaption(wchar_t const*) ()
   from /home/ubuntu/irrlicht/lib/Linux/libIrrlichtMt.so.1.9.0.13
#6  0x00005555555597e6 in main ()
```

Issue #274, opened 2024-01-05
ISceneNode::removeChild() is slow


Desour:
I've profiled Wuzzy's shadow forest game, which has many enemies that shoot with particles on you, and causes 4fps or so on my machine, even on low view range. I've used `RelWithDebInfo`. Here's a screenshot of the `Game::run()` part:
![shadowforest_rlsdbg](https://github.com/minetest/irrlicht/assets/7613443/ed03e7f9-7dd9-4819-83ca-841b3b39cf01)

Here's the implementation of `ISceneNode::removeChild()`, it iterates through all scene node childs:
https://github.com/minetest/irrlicht/blob/fb4ee6ac930c16a63127427a817593d81e218349/include/ISceneNode.h#L290-L303

This needs to be fixed for better particle performance.

SmallJoker:
The sorted `std::set` might perform better for loopkups (binary search?). Or perhaps `std::vector` if the compiler is smart enough to parallelize the lookup code in `std::find`.

appgurueu:
> The sorted `std::set` might perform better for loopkups (binary search?). Or perhaps `std::vector` if the compiler is smart enough to parallelize the lookup code in `std::find` (e.g. by using SIMD).

I don't think we need any sorting here. An unordered set would work, but is unnecessary. We have a doubly linked list and that allows O(1) removal - we just need the ISceneNode to store an iterator to its position in the list.

SmallJoker:
@appgurueu How would you seamlessly update those iterators when one child is removed? Storing iterators is in general bad practice as they may get outdated. If not now, someone will surely forget to update them in a future change.

appgurueu:
> @appgurueu How would you seamlessly update those iterators when one child is removed?

That shouldn't be necessary, a doubly linked list doesn't invalidate iterators.

> Storing iterators is in general bad practice as they may get outdated.

Doubly linked lists are special. Pretty much the only reason to use them is if you will be storing iterators to do object removal somewhere within the list in O(1) time.

> If not now, someone will surely forget to update them in a future change.

I would have liked to believe that this would be well isolated to the addChild and removeChild methods, but ASan seems to be proving me wrong.

Desour:
If in the future we want to replace the containers again or so, please note that iterating through also needs to be fast, because of functions like `OnAnimate` and `OnRegisterSceneNode`, which currently take about 1/8 client step cpu time that they mostly spend on iterating.

PR #275, opened 2024-01-05
Optimize scene node child removal to constant time

appgurueu:
An attempt at fixing #274, briefly tested with Shadow Forest, but not profiled.

This relies on two invariants in `removeChild`:

* If a node is a child of another node, its parent property is set appropriately. (And if a node isn't a child of another node, it's parent property **must** be different from that other node.)
* If a node is a child of another node, the iterator points to the correct list node.

You can experimentally verify that these seem to hold by replacing the body of `removeChild` with the less efficient:

```cpp
ISceneNodeList::iterator it = Children.begin();
for (; it != Children.end(); ++it)
	if ((*it) == child)
	{
		assert(child->Parent == this);
		assert(it == child->Iterator);
		child->Iterator = std::nullopt;
		child->Parent = 0;
		child->drop();
		Children.erase(it);
		return true;
	}

assert(child->Parent != this);
return false;
```

The `assert`s should not be tripped.

### DIFF
--- a/include/ISceneNode.h
+++ b/include/ISceneNode.h
@@ -277,8 +277,9 @@ namespace scene
 				child->grab();
 				child->remove(); // remove from old parent
 				Children.push_back(child);
-				child->Iterator = Children.end();
-				(*child->Iterator)--;
+				// Note: This iterator is not invalidated until we erase it.
+				child->ThisIterator = Children.end();
+				--(*child->ThisIterator);
 				child->Parent = this;
 			}
 		}
@@ -294,8 +295,9 @@ namespace scene
 			if (child->Parent != this)
 				return false;
 
-			auto it = child->Iterator.value();
-			child->Iterator = std::nullopt;
+			// The iterator must be set since the parent is not null.
+			auto it = *child->ThisIterator;
+			child->ThisIterator = std::nullopt;
 			child->Parent = nullptr;
 			child->drop();
 			Children.erase(it);
@@ -311,7 +313,7 @@ namespace scene
 		{
 			for (auto &child : Children) {
 				child->Parent = nullptr;
-				child->Iterator = std::nullopt;
+				child->ThisIterator = std::nullopt;
 				child->drop();
 			}
 			Children.clear();
@@ -617,10 +619,11 @@ namespace scene
 		//! Pointer to the parent
 		ISceneNode* Parent;
 
-		std::optional<ISceneNodeList::iterator> Iterator;
-
 		//! List of all children of this node
 		std::list<ISceneNode*> Children;
+
+		//! Iterator pointing to this node in the parent's child list.
+		std::optional<ISceneNodeList::iterator> ThisIterator;
 
 		//! Pointer to the scene manager
 		ISceneManager* SceneManager;
--- a/include/ISceneNode.h
+++ b/include/ISceneNode.h
@@ -15,6 +15,7 @@
 #include "matrix4.h"
 #include "IAttributes.h"
 #include <list>
+#include <optional>
 
 namespace irr
 {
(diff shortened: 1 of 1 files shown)


lhofhansl:
I know we discussed on the issue, but still wanted to ask:
1. Do we need to keep the insert order?
2. If not, can we perf-compare this with just using an unordered_set (assuming children are unique)?

include/ISceneNode.h
```diff
@@ -621,6 +617,8 @@ namespace scene
 		//! Pointer to the parent
 		ISceneNode* Parent;
 
+		std::optional<ISceneNodeList::iterator> Iterator;
```
SmallJoker, line 620: How about `ChildIterator`? Logically it would also make more sense to have to below `Children`.
Also a quick documentation like with all other members would be nice.

include/ISceneNode.h
```diff
@@ -276,30 +277,29 @@ namespace scene
 				child->grab();
 				child->remove(); // remove from old parent
 				Children.push_back(child);
+				child->Iterator = Children.end();
+				(*child->Iterator)--;
```
SmallJoker, line 281: It does make sense to me that the iterator never changes, but is this guaranteed by the C++ spec?

Desour: See "Iterator Invalidation" table here: https://en.cppreference.com/w/cpp/container

Btw., prefer pre-increment (doesn't need to copy the iterator):
```suggestion
				--(*child->Iterator);
```

include/ISceneNode.h
```diff
@@ -276,30 +277,29 @@ namespace scene
...
-					(*it)->Parent = 0;
-					(*it)->drop();
-					Children.erase(it);
-					return true;
-				}
+			if (child->Parent != this)
+				return false;
 
-			return false;
+			auto it = child->Iterator.value();
```
Desour, line 297: Maybe add a comment here saying that this will never throw an exception because Parent is not null.

Desour:
Tested, works.

appgurueu:
> Do we need to keep the insert order?

I'm not sure. I can imagine keeping the insertion order to be useful, for example when scene nodes need to be drawn in a particular order due to depth sorting. I don't know whether our current implementation relies on this however, since entities, particles, and mapblocks are all depth-sorted by Minetest.
 
> If not, can we perf-compare this with just using an unordered_set (assuming children are unique)?

We probably could, but I'm not sure an unordered set is a good idea.

Pros:

* *Maybe* slightly lower constant factors due to reduced allocator overhead.
* Probably slightly easier to maintain long-term.

Cons:

* Should it be deemed a worthwhile optimization, constant factors could also be lowered by giving scene nodes "prev" and "next" pointers, effectively making them list nodes. Since the scene nodes already need to be allocated, this should pretty much get rid of the allocator overhead.
* Larger change.
* Does not preserve order.
* Higher risk of breaking something.
* Unsure about shrinking / rehashing. If you only erase children, they [aren't even allowed](https://en.cppreference.com/w/cpp/container/unordered_set/erase) to do this. So if you go from 1000 children to 10, you will still have memory for 1000 children allocated (unless you manually call `rehash`, but you'd have to be careful to do that at the right time to not wreck performance).

> Just seems easy to accidentally break in the future.

As said on the issue, I believe this is well encapsulated by the add / remove child methods of `ISceneNode`. If you want further encapsulation, I can stick `Children` and `ChildIterator` in their own tiny "class" and make them `private`, only allowing access through methods which ensure consistency (though I feel that would be overengineered).

I also don't expect this (the children management in particular) to be touched often, looking at [the history](https://github.com/minetest/irrlicht/commits/master/include/ISceneNode.h).

include/ISceneNode.h
appgurueu: I think I'll add a comment and change this from `x.value()` to `*x`, given that Irrlicht is compiled without exceptions.

include/ISceneNode.h
appgurueu: I've opted for `ThisIterator`, given that this isn't the iterator of some child, but of this node in the parent's child list.

include/ISceneNode.h
```diff
@@ -276,30 +277,31 @@ namespace scene
...
-					(*it)->drop();
-					Children.erase(it);
-					return true;
-				}
+			if (child->Parent != this)
+				return false;
 
-			return false;
+			// The iterator must be set since the parent is not null.
+			auto it = *child->ThisIterator;
```
sfan5, line 299: prefer `.value()` here since that checks

appgurueu: I would use that, but don't we still compile Irrlicht with exceptions disabled? Will it `abort` if the check fails?

sfan5: :shrug:
you could also put `_IRR_DEBUG_BREAK_IF(!child->ThisIterator.has_value())` here (which we'll convert to an assert one day)

include/ISceneNode.h
```diff
@@ -276,30 +277,32 @@ namespace scene
 				child->grab();
 				child->remove(); // remove from old parent
 				Children.push_back(child);
+				// Note: This iterator is not invalidated until we erase it.
+				child->ThisIterator = Children.end();
```
savilli, line 281: `child->ThisIterator = std::prev(Children.end());` or `child->ThisIterator = Children.insert(Children.end(), child);`

include/ISceneNode.h
```diff
@@ -624,6 +623,9 @@ namespace scene
 		//! List of all children of this node
 		std::list<ISceneNode*> Children;
 
+		//! Iterator pointing to this node in the parent's child list.
+		std::optional<ISceneNodeList::iterator> ThisIterator;
```
savilli, line 627: Maybe put it right after `Parent` to show that they're essentially related.

savilli (approved):

include/ISceneNode.h
appgurueu: Thanks for the suggestion!

include/ISceneNode.h
appgurueu: I moved Parent right after this, so this is now between Parent and Children, the two fields its related to (the parent's children and this node's parent). I think that's a decent compromise.

sfan5 (approved):
tested a bit, didn't crash

PR #276, opened 2024-01-16
GL loader work

sfan5:
todo (not necessarily all here):
- [x] make extpointer mandatory
- [ ] figure out which of the changes broke world rendering on gles2
- [x] Drop unused exts from `COGLESCoreExtensionHandler`
- [ ] Make `COGLES2ExtensionHandler` use loader
- [ ] Make `COGLES1ExtensionHandler` use loader
- [ ] Drop unused exts and functions from `COpenGLExtensionHandler` (big oof)
- [ ] Make `COpenGLExtensionHandler` use loader
- [ ] Identify and migrate other GL calls
- [ ] Deal with `glext.h` ?!
- [ ] macOS support should automatically start working at this point


sfan5:
@numberZero thoughts?

SmallJoker:
(after requested in IRC)

I can confirm that OpenGL 2 still works (also with shaders) on Ubuntu/Linux, 64-bit. SDL2 / OpenGL 3 / OpenGL ES support not enabled.

lhofhansl:
I can confirm that OpenGL (4.6) still works on Linux, with shaders. FSAA also still works.

sfan5:
bad news: there's something wrong with either this PR or the unified GLES2 driver on Android:
test

grorp:
isn't this what it usually looks like when the Irrlicht shaders are missing?

sfan5:
Yes, however I'm reasonably sure that they aren't missing. The assets ZIP has them at least.

sfan5:
this is what renderdoc has to say about it:
grafik

sfan5:
Fixed it.

PR #278, opened 2024-01-16
Undefine SDL2 DirectFB video driver

okias:
1. we don't need it
2. it's dropped in SDL 3
3. it breaks compilation on Alpine and postmarketOS with SDL2 enabled (and the bug never going to get fixed on SDL2 side)

### DIFF
--- a/source/Irrlicht/CIrrDeviceSDL.h
+++ b/source/Irrlicht/CIrrDeviceSDL.h
@@ -17,6 +17,10 @@
 #endif
 
 #include <SDL.h>
+// DirectFB is removed in SDL3, thou distribution as Alpine currently ships SDL2
+// with enabled DirectFB, but requiring another fix at a top of SDL2.
+// We don't need DirectFB in Irrlicht/Minetest, so simply disable it here to prevent issues.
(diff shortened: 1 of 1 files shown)


okias:
It's a bit weird place, thou:
 - https://github.com/libsdl-org/SDL/pull/6084#issuecomment-1427222374
 - https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/29216/diffs
 
I got inspired by the second link. This removed the support for the Irrlicht, while keeping software really using DirectFB in Alpine having SDL2 support available.

In general it could be patched in Alpine or disable DirectFB in SDL2 https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/59109

sfan5:
Personally I would prefer if the the distribution would patch this when needed but also this single line change doesn't hurt and has no downsides.
I'll ask someone else to decide.

SmallJoker:
My two cents: this line is OK for me but there should be a comment explaining *why* this is needed and *when* it can be removed in the future (e.g. forcing SDL3 dependency).

PR #282, opened 2024-01-20
Specify dependencies on object library targets

JosiahWI:
The minimum CMake version was updated to 3.12 in #270. This cleans up the build using the feature to use target_link_libraries on object library targets that was added in CMake 3.12.

It has been tested on Linux with OpenGL. I will test it with more settings and on more systems as the work progresses.

### DIFF
--- a/source/Irrlicht/CMakeLists.txt
+++ b/source/Irrlicht/CMakeLists.txt
@@ -297,42 +297,10 @@ elseif(NOT USE_SDL2)
 	endif()
 endif()
 
-set(link_includes
-	"${PROJECT_SOURCE_DIR}/include"
-	"${CMAKE_CURRENT_SOURCE_DIR}"
-
-	"${ZLIB_INCLUDE_DIR}"
-	"${JPEG_INCLUDE_DIR}"
-	"${PNG_INCLUDE_DIR}"
-	"$<$<BOOL:${USE_SDL2}>:${SDL2_INCLUDE_DIRS}>"
-
-	${OPENGL_INCLUDE_DIR}
-	${OPENGLES2_INCLUDE_DIR}
-	${EGL_INCLUDE_DIR}
-
-	"$<$<PLATFORM_ID:Android>:${ANDROID_NDK}/sources/android/native_app_glue>"
-	"$<$<BOOL:${USE_X11}>:${X11_INCLUDE_DIR}>"
-)
-
-set(link_libs
-	"${ZLIB_LIBRARY}"
-	"${JPEG_LIBRARY}"
-	"${PNG_LIBRARY}"
-	"$<$<BOOL:${USE_SDL2}>:${SDL2_LIBRARIES}>"
-
-	"$<$<BOOL:${OPENGL_DIRECT_LINK}>:${OPENGL_LIBRARIES}>"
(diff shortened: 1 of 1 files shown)


source/Irrlicht/CMakeLists.txt
```diff
@@ -529,11 +537,27 @@ target_include_directories(IrrlichtMt
...
 
-target_link_libraries(IrrlichtMt PRIVATE ${link_libs})
+target_link_libraries(IrrlichtMt
+	PRIVATE
+		# FIXME this will not propogate dependencies to IrrlichtMt
+		# dependants, but I do not want to clutter the export with
+		# object libraries, either.
+		<$BUILD_INTERFACE:IRROBJ>
+		<$BUILD_INTERFACE:IRROTHEROBJ>
+		<$BUILD_INTERFACE:IRRVIDEOOBJ>
```
JosiahWI, line 549: This is not a portable solution. I will probably end up including these in the export for the time being.

numberZero:
Can’t you just replace `target_sources` with linking to the object libraries?

JosiahWI:
> Can’t you just replace `target_sources` with linking to the object libraries?

I'm planning to try that. I expect it will work.

source/Irrlicht/CMakeLists.txt
```diff
@@ -406,6 +374,23 @@ add_library(IRRVIDEOOBJ OBJECT
...
+		"$<$<BOOL:${OPENGLES_DIRECT_LINK}>:${OPENGLES_LIBRARY}>"
+		"$<$<BOOL:${OPENGLES2_DIRECT_LINK}>:${OPENGLES2_LIBRARIES}>"
+		"${EGL_LIBRARY}"
+		JPEG::JPEG
+	PRIVATE
+		PNG::PNG
+)
+
+if(USE_X11)
+	target_include_directories(IRRVIDEOOBJ PUBLIC "${X11_INCLUDE_DIR}")
```
sfan5, line 391: this could be a generator expression instead

source/Irrlicht/CMakeLists.txt
```diff
@@ -406,6 +374,23 @@ add_library(IRRVIDEOOBJ OBJECT
...
+target_link_libraries(IRRVIDEOOBJ
+	PUBLIC
+		# TODO Create imported targets if they do not already exist,
+		# and only link the correct one for the enabled backend.
+		"$<$<BOOL:${OPENGL_DIRECT_LINK}>:${OPENGL_LIBRARIES}>"
+		"$<$<BOOL:${OPENGLES_DIRECT_LINK}>:${OPENGLES_LIBRARY}>"
+		"$<$<BOOL:${OPENGLES2_DIRECT_LINK}>:${OPENGLES2_LIBRARIES}>"
+		"${EGL_LIBRARY}"
+		JPEG::JPEG
+	PRIVATE
```
sfan5, line 386: by which logic is public/private decided here?

source/Irrlicht/CMakeLists.txt
JosiahWI: That will be obsolete, though, once I change it to use a target.

source/Irrlicht/CMakeLists.txt
JosiahWI: I usually base it on whether the dependency is included from a public header. I will double check all of these before taking it out of draft.

PR #285, opened 2024-02-07
[SDL] Improve handling of IMEs

y5nw:
This PR sets the text input rectangle when text input is used. It allows (among other things) IMEs to position the text input candidate list correctly.

The screenshots below demonstrate the difference. In both screenshots, the "World name" input box is focused.

Before:

After:

This also (at least partially) fixes minetest/minetest#13473 (dead keys and compose keys do seem to work under IBus for me now) by not restarting text input when this is not necessary.

### DIFF
--- a/source/Irrlicht/CIrrDeviceSDL.cpp
+++ b/source/Irrlicht/CIrrDeviceSDL.cpp
@@ -219,17 +219,29 @@ void CIrrDeviceSDL::resetReceiveTextInputEvents() {
 	gui::IGUIElement *elem = GUIEnvironment->getFocus();
 	if (elem && elem->acceptsIME())
 	{
-		core::rect<s32> pos = elem->getAbsolutePosition();
-		SDL_Rect rect;
-		rect.x = pos.UpperLeftCorner.X;
-		rect.y = pos.UpperLeftCorner.Y;
-		rect.w = pos.getWidth();
-		rect.h = pos.getHeight();
-		SDL_SetTextInputRect(&rect);
-		SDL_StartTextInput();
+		core::rect<s32> *pos = new core::rect<s32>(elem->getAbsolutePosition());
+		if (lastElemPos == NULL || *lastElemPos != *pos)
+		{
+			if (lastElemPos != NULL)
+				delete lastElemPos;
+			lastElemPos = pos;
+			SDL_Rect rect;
+			rect.x = pos->UpperLeftCorner.X;
+			rect.y = pos->UpperLeftCorner.Y;
+			rect.w = pos->getWidth();
+			rect.h = pos->getHeight();
+			SDL_SetTextInputRect(&rect);
+			SDL_StartTextInput();
+		}
+		else
+		{
+			delete pos;
+		}
 	}
 	else
 	{
+		delete lastElemPos;
+		lastElemPos = NULL;
 		SDL_StopTextInput();
 	}
 }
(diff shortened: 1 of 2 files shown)


source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -218,9 +218,20 @@ int CIrrDeviceSDL::findCharToPassToIrrlicht(int assumedChar, EKEY_CODE key) {
...
 	if (elem && elem->acceptsIME())
+	{
 		SDL_StartTextInput();
+		core::rect<s32> pos = elem->getAbsolutePosition();
+		SDL_Rect rect;
+		rect.x = pos.UpperLeftCorner.X;
+		rect.y = pos.UpperLeftCorner.Y;
+		rect.w = pos.LowerRightCorner.X - rect.x;
+		rect.h = pos.LowerRightCorner.Y - rect.y;
+		SDL_SetTextInputRect(&rect);
```
appgurueu, line 229: This should ideally be called before `SDL_StartTextInput` as per the [SDL docs](https://wiki.libsdl.org/SDL2/SDL_SetTextInputRect).

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -218,9 +218,20 @@ int CIrrDeviceSDL::findCharToPassToIrrlicht(int assumedChar, EKEY_CODE key) {
 void CIrrDeviceSDL::resetReceiveTextInputEvents() {
 	gui::IGUIElement *elem = GUIEnvironment->getFocus();
 	if (elem && elem->acceptsIME())
+	{
 		SDL_StartTextInput();
+		core::rect<s32> pos = elem->getAbsolutePosition();
+		SDL_Rect rect;
+		rect.x = pos.UpperLeftCorner.X;
+		rect.y = pos.UpperLeftCorner.Y;
+		rect.w = pos.LowerRightCorner.X - rect.x;
```
appgurueu, line 227: Use [`pos.getWidth()`](https://irrlicht.sourceforge.io/docu/classirr_1_1core_1_1rect.html#ae313a3674095733348d5e8bcb75c49b4) here

source/Irrlicht/CIrrDeviceSDL.cpp
```diff
@@ -218,9 +218,20 @@ int CIrrDeviceSDL::findCharToPassToIrrlicht(int assumedChar, EKEY_CODE key) {
...
+		rect.h = pos.LowerRightCorner.Y - rect.y;
```
appgurueu, line 228: Use `pos.getHeight()` here

appgurueu (changes requested):
Thanks for the PR. I have some comments. The first two are minor style issues, the last one is probably important.

source/Irrlicht/CIrrDeviceSDL.cpp
y5nw: Thanks. It seems like I went through the relevant SDL pages faster than I should.

source/Irrlicht/CIrrDeviceSDL.cpp
y5nw: Thanks for pointing this out. I mostly took the code snippet from its `CIrrDeviceLinux` counterpart and mostly expected this to be a quick fix.

y5nw:
Actually I think I can fix another (semi-related) bug while I am at it.

source/Irrlicht/CIrrDeviceSDL.h
```diff
@@ -303,6 +303,8 @@ namespace irr
 
 		bool Resizable;
 
+		core::rect<s32> *lastElemPos;
```
appgurueu, line 306: Why a pointer? Just use `std::optional` here, saves you the manual memory management headache.

y5nw: Thanks for the tip. (I am more used to C than C++ and was not aware of `std::optional`.)

sfan5 (approved):

PR #287, opened 2024-02-11
Allow VBO to be used for meshes all the way down to 4 vertices.

paradust7:
I used linux perf and GPUView (Windows) to examine rendering performance. I discovered in both cases that Minetest was CPU-bound, with the GPU barely being used. In particular, the CPU was spending most of its time in memory operations (CPU to GPU). I measured the transfer rate to be around 30GB/sec sustained.

Looking at the code, I discovered that the vast majority (80%) of map block meshes are under 500 vertices, and were therefore not being stored in VBOs, because of this arbitrary limit. They were being retransmitted in full every frame.

This boosts worst-case FPS (in Mineclonia) around 45% on my hardware, from 65fps to 95fps. (with default rendering distance)

### DIFF
--- a/source/Irrlicht/CNullDriver.cpp
+++ b/source/Irrlicht/CNullDriver.cpp
@@ -45,7 +45,7 @@ IImageWriter* createImageWriterPNG();
 //! constructor
 CNullDriver::CNullDriver(io::IFileSystem* io, const core::dimension2d<u32>& screenSize)
 	: SharedRenderTarget(0), CurrentRenderTarget(0), CurrentRenderTargetSize(0, 0), FileSystem(io), MeshManipulator(0),
-	ViewPort(0, 0, 0, 0), ScreenSize(screenSize), PrimitivesDrawn(0), MinVertexCountForVBO(500),
+	ViewPort(0, 0, 0, 0), ScreenSize(screenSize), PrimitivesDrawn(0), MinVertexCountForVBO(4),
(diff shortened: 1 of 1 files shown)


grorp:
There's this related code in Minetest:

https://github.com/minetest/minetest/blob/eb52a149a09dd8811d5defc4fe85af5bcc84dba7/src/client/clientlauncher.cpp#L117-L121

sfan5:
Yeah we can change this in MT, no need to adjust Irrlicht defaults.

I checked the history of the commented out call and it has been commented out since the very first commits it appeared in (minetest/minetest@4e249fb3fbf75f0359758760d88e22aa5b14533c)...

lhofhansl:
@paradust7 Did you also try to set `client_mesh_chunk`. Default is 1, changing to larger values reduces the number of meshes in favor of fewer larger ones.

paradust7:
> @paradust7 Did you also try to set `client_mesh_chunk`. Default is 1, changing to larger values reduces the number of meshes in favor of fewer larger ones.

I haven't, but as long as everything is in VBO, I don't see there being any extra benefit to joining meshes (and there might be downsides).

lhofhansl:
> I haven't, but as long as everything is in VBO, I don't see there being any extra benefit to joining meshes (and there might be downsides).

There's the number of drawcalls. I still find - even with the MT patch applied - that setting `client_mesh_chunk` to 8 gets my FPS from 20 to 60 FPS. (Older mobile NVidia card, viewing_range = 1000)

PR #288, opened 2024-02-11
Fix undefined behaviors

fuzun:
This merge request fixes two undefined behaviors found by using Undefined Behavior Sanitizer.

### DIFF
--- a/source/Irrlicht/CImageLoaderTGA.cpp
+++ b/source/Irrlicht/CImageLoaderTGA.cpp
@@ -10,6 +10,7 @@
 #include "CImage.h"
 #include "irrString.h"
 
+#define MAX(x, y) (((x) > (y)) ? (x) : (y))
 
 namespace irr
 {
@@ -138,7 +139,8 @@ IImage* CImageLoaderTGA::loadImage(io::IReadFile* file) const
 	if (header.ColorMapType)
 	{
 		// Create 32 bit palette
-		const irr::u16 paletteSize = core::max_((u16)256, header.ColorMapLength);	// ColorMapLength can lie, but so far we only use palette for 8-bit, so ensure it has 256 entries
+		// `core::max_()` is not used here because it takes its inputs as references. Since `header` is packed, use the macro `MAX()` instead:
+		const irr::u16 paletteSize = MAX((u16)256u, header.ColorMapLength); // ColorMapLength can lie, but so far we only use palette for 8-bit, so ensure it has 256 entries
 		palette = new u32[paletteSize];
 
 		if( paletteSize > header.ColorMapLength )
--- a/source/Irrlicht/CIrrDeviceSDL.h
+++ b/source/Irrlicht/CIrrDeviceSDL.h
@@ -260,7 +260,7 @@ namespace irr
 				}
 			};
 			std::vector<std::unique_ptr<SDL_Cursor, CursorDeleter>> Cursors;
-			gui::ECURSOR_ICON ActiveIcon;
+			gui::ECURSOR_ICON ActiveIcon = gui::ECURSOR_ICON::ECI_NORMAL;
 		};
 
 	private:


source/Irrlicht/CImageLoaderTGA.cpp
```diff
@@ -138,7 +139,7 @@ IImage* CImageLoaderTGA::loadImage(io::IReadFile* file) const
 	if (header.ColorMapType)
 	{
 		// Create 32 bit palette
-		const irr::u16 paletteSize = core::max_((u16)256, header.ColorMapLength);	// ColorMapLength can lie, but so far we only use palette for 8-bit, so ensure it has 256 entries
+		const irr::u16 paletteSize = MAX((u16)256u, (u16)(header.ColorMapType)); // ColorMapLength can lie, but so far we only use palette for 8-bit, so ensure it has 256 entries
```
appgurueu, line 142: I'd simply use `std::max` (`#include <algorithm>`) for this. You should be able to write `std::max<u16>(256, header.ColorMapType)` and have it work.

fuzun: According to here https://en.cppreference.com/w/cpp/algorithm/max, `std::max` is defined as:

```
const T& max( const T& a, const T& b );
```

This means that we would have the same problem if we use `std::max`. Or, am I missing something here?

appgurueu: `ColorMapLength` is already `u16`. Why do you want to cast that to `u16`?

`256` is in principle an `int` literal. But when we pass it to a function which expects `u16`, it will be converted; since it fits both in an `int` (without the sign bit being set) and in `u16`, all is fine. I don't see the issue here. Though feel free to keep the `u` suffix if you prefer that.

(Also, I just see that `ColorMapLength` mutated to `ColorMapType`. Please fix that.)

fuzun: > (Also, I just see that ColorMapLength mutated to ColorMapType. Please fix that.)

Thanks for commenting this! I missed that part, will fix it now.

> ColorMapLength is already u16. Why do you want to cast that to u16?

Yeah it is, no casting is needed, I mistakenly referred to `ColorMapType` there instead, which happened during synchronizing my development branch and the testing branch. 

>  I don't see the issue here. Though feel free to keep the u suffix if you prefer that.

The problem seems that since references are implemented using pointers, we can not pass a member of a packed struct because the address may not be aligned.

appgurueu: Ah I see, I wasn't aware that the struct was packed (frankly, I would say this is hacky and error-prone and should be refactored eventually, but not in this PR). I would recommend explaining this in a comment.

appgurueu (approved):
Looks good. (Though as said a comment would be good to have.)

source/Irrlicht/CImageLoaderTGA.cpp
fuzun: I added a comment, thanks for reviewing!

appgurueu: Thank you for the testing & PR!

source/Irrlicht/CImageLoaderTGA.cpp
```diff
@@ -138,7 +139,8 @@ IImage* CImageLoaderTGA::loadImage(io::IReadFile* file) const
 	if (header.ColorMapType)
 	{
 		// Create 32 bit palette
-		const irr::u16 paletteSize = core::max_((u16)256, header.ColorMapLength);	// ColorMapLength can lie, but so far we only use palette for 8-bit, so ensure it has 256 entries
+		// `core::max_()` is not used here because it takes its inputs as references. Since `header` is packed, use the macro `MAX()` instead:
```
sfan5, line 142: what about `std::max`?

appgurueu: > what about `std::max`?

Same issue, see https://github.com/minetest/irrlicht/pull/288#discussion_r1485651314. Since `std::max` is technically a function, it seems the undefined behavior would technically persist as we pass on a reference to an unaligned struct member.

The macro is one solution, another would be to first copy `header.ColorMapLength` to a local variable. I can live with either.

sfan5: That's unfortunate.
Now since this seems like such a common trap I am left wondering: is it *actually* UB if the template never takes the address of the argument?

fuzun: That's a good point, but I think we can not make sure of that. As far as I know, references always make use of pointers in the background.

appgurueu: > As far as I know, references always make use of pointers in the background.

I don't think they have to; the C++ compiler will most probably optimize references out in many cases on higher optimization levels, especially const references to types which are always cheaper to pass by value (as is the case here) if it can inline the function (as is the case for `std::max`).

But you're right that we most probably can't rely on this, especially as the compiler needn't optimize on `-O0`.

PR #289, opened 2024-02-17
Remove support for GL_POINT_SMOOTH and GL_LINE_SMOOTH antialiasing

HybridDog:
These antialiasing techniques have been removed in OpenGL 3.1, they were often executed by the CPU, and Minetest does not use them. The OpenGL wiki recommends that we do not use this functionality in our program. https://www.khronos.org/opengl/wiki/Multisampling#Smooth_antialiasing

### DIFF
--- a/source/Irrlicht/COGLESDriver.cpp
+++ b/source/Irrlicht/COGLESDriver.cpp
@@ -1796,7 +1796,6 @@ void COGLES1Driver::setBasicRenderStates(const SMaterial& material, const SMater
 	// Anti aliasing
 	if (resetAllRenderStates || lastmaterial.AntiAliasing != material.AntiAliasing)
 	{
-//		if (FeatureAvailable[IRR_ARB_multisample]) {
 		if (material.AntiAliasing & EAAM_ALPHA_TO_COVERAGE)
 			glEnable(GL_SAMPLE_ALPHA_TO_COVERAGE);
 		else if (lastmaterial.AntiAliasing & EAAM_ALPHA_TO_COVERAGE)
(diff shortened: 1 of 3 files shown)


source/Irrlicht/COGLESDriver.cpp
```diff
@@ -1798,33 +1796,16 @@ void COGLES1Driver::setBasicRenderStates(const SMaterial& material, const SMater
 	// Anti aliasing
 	if (resetAllRenderStates || lastmaterial.AntiAliasing != material.AntiAliasing)
 	{
-//		if (FeatureAvailable[IRR_ARB_multisample])
-		{
-			if (material.AntiAliasing & EAAM_ALPHA_TO_COVERAGE)
-				glEnable(GL_SAMPLE_ALPHA_TO_COVERAGE);
-			else if (lastmaterial.AntiAliasing & EAAM_ALPHA_TO_COVERAGE)
-				glDisable(GL_SAMPLE_ALPHA_TO_COVERAGE);
+//		if (FeatureAvailable[IRR_ARB_multisample]) {
```
sfan5, line 1799:
```suggestion
```

lhofhansl:
Tested. Works. Including FSAA. (Linux)

source/Irrlicht/COpenGLDriver.cpp
```diff
@@ -2557,46 +2555,29 @@ void COpenGLDriver::setBasicRenderStates(const SMaterial& material, const SMater
 	}
 
 	// Anti aliasing
-	if (resetAllRenderStates || lastmaterial.AntiAliasing != material.AntiAliasing)
-	{
-		if (FeatureAvailable[IRR_ARB_multisample])
+	if (resetAllRenderStates
+			|| lastmaterial.AntiAliasing != material.AntiAliasing
+			&& FeatureAvailable[IRR_ARB_multisample]) {
```
sfan5, line 2560: on second look this is actually broken
needs to be `(A || B) && C`

HybridDog: You're right; thanks.

PR #290, opened 2024-02-17
Upgrade actions in CI

wsor4035:
Add compact, short information about your PR for easier understanding:

- upgrades all the actions use in workflows to resolve warnings
- created per @sfan5 request in https://github.com/minetest/minetest/pull/14377#issuecomment-1949941967

## To do

This PR is a Work in Progress

- [x] verify no warnings when actions finish

## Outside the scope of just upgrading actions

lukka/run-vcpkg@v7 - https://github.com/minetest/minetest/pull/14377#discussion_r1493006875

## How to test

use your eyes to check actions reports

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -258,7 +258,7 @@ jobs:
       - name: Upload Artifact
         uses: actions/upload-artifact@v4
         with:
-          name: msvc-${{ matrix.config.arch }}
+          name: msvc-${{ matrix.config.arch }}-${{matrix.sdl.use}}
           path: artifact/
 
   android:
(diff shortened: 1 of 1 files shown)


wsor4035:
two failing appear to be unrelated to changes made in this pr

sfan5:
Not unrelated but this bug was previously hidden.
You can change the msvc workflows to only upload the sdl artifacts for example.

wsor4035:
> Not unrelated but this bug was previously hidden. You can change the msvc workflows to only upload the sdl artifacts for example.

could this also be solved by changing https://github.com/minetest/irrlicht/blob/master/.github/workflows/build.yml#L261 to be `msvc-${{ matrix.config.arch }}-${{matrix.sdl.use}}` potentially? thoughts?

sfan5:
Yes. Either works.

wsor4035:
in theory fixed (see when tests finish in ... )

appgurueu (approved):

